PR #227, opened 2016-10-27
Updated the audio_playback.ipynb example

yunqu:
Changes include:

1. Updated the `audio_playback.ipynb` notebook in `notebooks/examples/` folder, with a faster preprocessing time and better visualization of frequency domain. Polished markdown comments. The script cells are reorganized, with required steps clearly stated.

2. Truncated the audio sample `pynq_welcome.pdm` from @schelleg. The sample length is reduced from 4 seconds to 3 seconds; the trailing (quiet) 1-second audio is removed. The first 3 seconds of the audio sample stays untouched. This is done to avoid confusion when displaying the frequency domain.


PeterOgden:
Are you able to rebase your commits on to the latest version of master? There is a lot of unrelated noise in this request.

PR #229, opened 2016-11-03
Improve performance of MMIO class

PeterOgden:
I've been working on rewriting some aspects of the code which are causing performance bottlenecks for various applications. First two targets are the MMIO class which was hampering my attempts for a pure python audio driver, and the Trace buffer parser which was taking nearly a minute to run in the example notebooks. I would appreciate someone checking I've not removed some vital piece of functionality in the process.

### DIFF
--- a/python/pynq/mmio.py
+++ b/python/pynq/mmio.py
@@ -55,8 +55,12 @@ class MMIO:
         The length in bytes of the address range.
     debug : bool
         Turn on debug mode if it is True.
+    mmap_file : file
+        Underlying file object for MMIO mapping
     mem : mmap
         An mmap object created when mapping files to memory.
+    array : numpy.ndarray
+        A numpy view of the mapped range for efficient assignment
     
     """
 
@@ -85,13 +89,7 @@ class MMIO:
         
         # Calculate base address offset w.r.t the base address
         self.virt_offset = base_addr - self.virt_base
-        
-        # Align the stop address with the words
-        stop = base_addr + length
-        if (stop % general_const.MMIO_WORD_MASK):
-            stop = (stop + general_const.MMIO_WORD_LENGTH) & \
-                    general_const.MMIO_WORD_MASK
-        
+
         # Storing the base address and length
         self.base_addr = base_addr
         self.length = length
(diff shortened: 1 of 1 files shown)


cathalmccabe:
Hi Peter, looks like the bounds checking is removed in the mmio. Was this intentional or have I missed it somewhere?

python/pynq/mmio.py
```diff
@@ -96,16 +97,18 @@ def __init__(self, base_addr, length = 4, debug = False):
...
-        self._debug('MMIO(address, size) = ({0}, {1} bytes).'.\
-                    format(hex(self.base_addr), hex(self.length)))
+        self._debug('MMIO(address, size) = ({0:x}, {1:x} bytes).',
+                    self.base_addr,self.length)
                 
         # Open file and mmap
-        f = os.open(general_const.MMIO_FILE_NAME, os.O_RDWR | os.O_SYNC)
-        self.mem = mmap.mmap(f, (self.length + self.virt_offset), 
+        self.f = os.open(general_const.MMIO_FILE_NAME, os.O_RDWR | os.O_SYNC)
+        self.mem = mmap.mmap(self.f, (self.length + self.virt_offset),
```
yunqu: Notice that self.f is only used once. Is it really necessary to keep this as an attribute?

python/pynq/mmio.py
```diff
@@ -38,6 +38,7 @@
 import mmap
 import math
 from . import general_const
+import numpy as np
```
yunqu: The use of numpy is kind of an overkill, since we do not do arithmetic over the data. I am not sure if we really need this.

python/pynq/mmio.py
PeterOgden: We ought to close the file when the MMIO object gets cleans up, hence why I turned it into an attribute. I forgot to check in the close method. Thoughts on an explicit `close` vs. using `__del__` welcome. I vote for using `__del__` as it isn't essential we always clean up, it would just be preferable to not have 10s or 100s of file descriptors pointing at `/dev/mem` while playing around with driver development.

python/pynq/mmio.py
PeterOgden: The argument behind using numpy is that it provides a way to view an mmapped file as an array of 32-bit ints which is essentially what we have rewritten in a custom way. I don't think it is possible to do the same using the array class. I experimented with various ways of reading and writing and numpy offered by far the best total performance, any additional features are just bonuses.

python/pynq/mmio.py
```diff
@@ -157,50 +156,38 @@ def write(self, offset, data):
...
-            self._debug('Writing 4 bytes to offset {0}: {1}'.\
-                        format(hex(offset), hex(data)))
-            mem.write(struct.pack('I', data))
+            self._debug('Writing 4 bytes to offset {0:x}: {1:x}',
+                        offset, data)
+            self.array[idx] = np.uint32(data)
+        elif type(data) is bytes:
+            length = len(data)
+            num_words = length >> 2
+            if (num_words << 2 != length):
```
yunqu: Line 171, please remove redundant parenthesis.

python/pynq/mmio.py
yunqu: I would go for __del__.

python/pynq/mmio.py
```diff
@@ -96,16 +97,18 @@ def __init__(self, base_addr, length = 4, debug = False):
         self.length = length
```
yunqu: For line 92, it seems that `stop` is no longer used, can we remove this variable completely?

PeterOgden:
I've submitted additional changes to fix the review comments and run autopep8 on the code. I've separated them out into two separate commits because of the noise that autopep8 introduces. I've also removed the tracebuffer commit from the pull request pending further review.

PR #231, opened 2016-11-10
Improved arduino analog voltage conversion

LucaCerina:
Hello,
testing the arduino analog with multiple analog channels I've noticed that read() [voltage conversion read] decreases its sampling frequency for multiple channels or increasing samples.
Looking in the code I've noticed that a floating point division is V_REF / 65536 is repeated for every read of every channel.
Moving it to the main before the mailbox command loop extensively increased the conversion performances.


LucaCerina:
done

yunqu:
@LucaCerina  Thanks for the changes you made. I just noticed that the binary file (arduino_analog.bin) file is not updated. Is it possible to update this file as well, so this pull request will be a complete one?

LucaCerina:
done

yunqu:
@LucaCerina  I guess this pull request will be redundant because of #234. So I will close this and only review #234 . Let me know if there is any problem.

PR #237, opened 2016-12-08
Doc update

cathalmccabe:
Update getting started (video guide added)
Update Creating overlays


python/pynq/mmio.py
```diff
@@ -38,10 +38,12 @@
 import mmap
 import math
 from . import general_const
+import numpy as np
```
yunqu: `import numpy as np` has to be in front of `from . import general_const`

PR #244, opened 2017-01-09
Make our scripts/makefile agnostic to the python version

PeterOgden:
This change also includes some scripts to source a version of
python from /etc/profile.d as this does not happen by default
when scripts are called directly from the init system

### DIFF
--- a/scripts/linux/3_pl_server.sh
+++ b/scripts/linux/3_pl_server.sh
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-rm -rf /home/xilinx/pynq/bitstream/.log
-python3.4 /home/xilinx/scripts/start_pl_server.py &
+if [ -z "$PYNQ_PYTHON" ]; then
+  PYNQ_PYTHON=python3.4
+fi
+
+if [ -f /etc/profile.d/python3.6.sh ]; then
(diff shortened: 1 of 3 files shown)


scripts/linux/3_pl_server.sh
```diff
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-rm -rf /home/xilinx/pynq/bitstream/.log
```
yunqu: @PeterOgden .log file is required in the PL class. If we run into any problem with the PL class socket, this file may not be deleted automatically. So I suggest we still keep `rm -rf .log` statement.

scripts/linux/4_boot_leds.sh
```diff
@@ -1,4 +1,11 @@
 #!/bin/bash
 
-rm -rf /home/xilinx/pynq/bitstream/.log
```
yunqu: @PeterOgden  This is fine. We don't have to delete it again since `3_pl_server.sh` already does it.

scripts/linux/3_pl_server.sh
PeterOgden: We already delete this file in start_pl_server.py. Are there any circumstances where we need to delete it again here?

PR #246, opened 2017-01-11
Add Interrupt Support to PYNQ

PeterOgden:
Adds support to PYNQ. Consists of several sub-units of work
 * Updated hardware design to include interrupts. Also added IOP/DDR interconnect (Parimal)
 * Add AXI interrupt controller drivers
 * Update PL class to keep track of interrupts and parse interrupt connections from overlay TCL files
 * Add user-facing `Interrupt` class which acts similarly to an `asyncio.Event` but relies on the hardware being cleared to reset it
 * Update Switch and Button classes to provide `asyncio` and blocking function calls
 * New example notebook to demonstrate the Button/Switch APIs


python/pynq/board/button.py
```diff
@@ -1,39 +1,43 @@
...
+#   OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
 #   ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 __author__ = "Giuseppe Natale"
 __copyright__ = "Copyright 2016, Xilinx"
 __email__ = "pynq_support@xilinx.com"
 
 
+import asyncio
+
```
yunqu: Pls remove the extra empty line.

python/pynq/pl.py
```diff
@@ -435,7 +494,6 @@ def __init__(self, bitfile_name):
         ----------
         bitfile_name : str
             The bitstream absolute path or name as a string.
-
```
yunqu: I think there should an empty line here?

python/pynq/interrupt_map.py
```diff
@@ -0,0 +1,139 @@
...
+        # Initialize result variables
+        self.intc_names = []
+        self.intc_parent = {}
+        self.concat_cells = {}
+        self.nets = []
+        self.pins = {}
+        self.intc_pins = {}
+        self.ps7_name = ""
+
+        # Key strings to search for in the TCL file
```
yunqu: The following string expression definitions seem too many. I would remove some of those and just use regex. For example, `create_bd_cell`. And I would not define expressions like "] $".

PeterOgden:
I would agree with the organisation. I think keeping the parsing and driver code as separate classes is a good idea. I contemplated moving the parsing code into the PL class as that is the sole user of that particular bit of logic, but it is fairly long and detracts from the main purpose of the class hence I separated it out. It could live as a separate function inside of the interrupt source file.

The user facing code boils down to the single class `Interrupt` and that is the only class brought into the pynq.driver namespace (soon to be pynq namespace). Separating out the Interrupt, InterruptController and UIO classes simplifies a lot of the implementation as the classes can be used polymorphically as parents of each other.

I propose moving all of the interrupt related code into a single interrupt.py at pynq level with all but the Interrupt class and a parsing function prefixed with an underscore to denote the private nature of the implementation details. The parsing function would be imported explicitly by pl.py and the toplevel __init__.py would only bring in `Interrupt`. Thoughts?

python/pynq/interrupt.py
```diff
@@ -0,0 +1,267 @@
...
+        self.mmio.write(0x0C, irqs)
+        if self.event_number:
+            self.parent.add_event(self, self.number)
+
+    def add_event(self, event, number):
+        """Registers an event against an interrupt line
+
+        Parameters
+        ----------
+            event : object
```
yunqu: The indentation here may cause the doc build flow to fail.

python/pynq/interrupt.py
```diff
@@ -0,0 +1,267 @@
...
+                Any object that provides a set method to notify of
+                an active interrupt
+            number : int
```
yunqu: Same indentation here.

python/pynq/interrupt.py
```diff
@@ -0,0 +1,267 @@
...
+                Interrupt number to register event against
+
+        When the interrupt is active, all events are signaled and the
```
yunqu: Better move this piece of note up before the section `Parameters`.

python/pynq/pl.py
```diff
@@ -152,6 +152,165 @@ def _get_gpio(tcl_name):
...
+        ---------
+        tcl_name : str
+            The tcl filename to parse. This is opened directly so should be
+            fully qualified
+
+        """
+        if not isinstance(tcl_name, str):
+            raise TypeError("tcl_name has to be a string")
+
+        # TODO: support nested hierarchies
```
yunqu: Remove?

python/pynq/pl.py
```diff
@@ -152,6 +152,165 @@ def _get_gpio(tcl_name):
...
+        self.intc_pins = {}
+        self.ps7_name = ""
+
+        # Key strings to search for in the TCL file
+        hier_pat = "create_hier_cell"
+        concat_pat = "create_bd_cell -type ip -vlnv xilinx.com:ip:xlconcat:2.1"
+        interrupt_pat = "create_bd_cell -type ip -vlnv xilinx.com:ip:axi_intc:4.1"
+        ps7_pat = "create_bd_cell -type ip -vlnv xilinx.com:ip:processing_system7:5.5"
+        prop_pat = "set_property -dict"
+        end_prop_pat = "] $"
```
yunqu: Seems like this variable is never used? I suggest to use python editors like pycharm to easily fix this kind of problem.

python/pynq/pl.py
```diff
@@ -152,6 +152,165 @@ def _get_gpio(tcl_name):
 
     return result
 
+class _InterruptMap:
```
yunqu: InterruptMap -> Interrupt_Map ?

PeterOgden: Wouldn't that violate the PEP-8 naming convention? I've addressed all other comments in a review commit.

PR #249, opened 2017-01-13
Arduino_lcd18 support

yunqu:
This pull request has the following changes:

1. Added Arduino LCD 18 support. Users are able to provide a small picture (160*128) in (jpeg/png/bmp/...) with a file path and show that picture on the LCD 18 screen. Users will also be able to draw lines, print characters, and animate the picture shown on the screen.
2. Updated the base.hdf file and rebuilt all the SDK projects. The hdf file is the latest one consistent with @PeterOgden 's bitstream.


yunqu:
For `video.py`, based on:
http://stackoverflow.com/questions/384759/pil-and-numpy
http://effbot.org/zone/pil-changes-116.htm

The best practice should be convert numpy array directly to image, instead of doing any additional conversion such as numpy array -> bytearray -> bytes -> image. 

Also removed `save_raw_as_jpeg()`, since it is not used anywhere in our codes, and is almost the same as `save_as_jpeg()`.

yunqu:
@schelleg  I have updated the API. Let me know if there is still any other issue pending.

yunqu:
@schelleg I have renamed all the files. Let me know if I need to do anything else. Once this pull request has been pushed onto github, I will make future pull requests feature-related.

schelleg:
We'll break this up into feature pull-requests and accept in pieces.

PR #251, opened 2017-01-17
minor fix in usb_wifi class

LucaCerina:
I modified the class as requested by @yunqu. Now an exception is raised if wifi device is not recognized (not compatible, not attached or driver loading failed)


Pynq-Z1/sdk/arduino_analog/src/arduino_analog.c
```diff
@@ -47,7 +47,7 @@
  * ----- --- -------- -----------------------------------------------
  * 1.00a pp  09/01/16 release
  * 1.00b yrq 09/06/16 adjust format, change log size
- * 1.00c lcc 11/10/16 voltage reference with single fl.point division
+ * 1.00c LcC 11/10/16 voltage reference with single fl.point division
```
yunqu: Is this change necessary? I don't think you need to change this file, right?

python/pynq/drivers/usb_wifi.py
```diff
@@ -71,7 +71,11 @@ def __init__(self):
             if m:
                 if m.group(2) != 'lo' and m.group(2) != 'eth0':
                     self.wifi_port = m.group(2)
-        print(self.wifi_port)
+        if(hasattr(self,'wifi_port')):
+            print(self.wifi_port)
```
yunqu: I think it is better to have `self.wifi_port` always defined, even if it is None. 

Also, maybe we need to define another parameter `verbose`, and make its default value `True`; you will only print message when verbose is `True`, otherwise we don't print out anything.
i.e. something similar to: 
```python
__init__(self, verbose=True)
```

yunqu:
@LucaCerina  A few comments:

1. Can you rebase your change, so only 1 commit will be in this pull request?

2. Can you pull against the branch image_2017_01?

LucaCerina:
ok, I will fix it

PR #268, opened 2017-01-25
base overlay - PMOD backwards compatible

schelleg:


Pynq-Z1/bitstream/base.tcl
```diff
@@ -176,7 +176,7 @@ proc create_hier_cell_mb3_timers_subsystem { parentCell nameHier } {
   variable script_folder
 
   if { $parentCell eq "" || $nameHier eq "" } {
-     catch {common::send_msg_id "BD_TCL-102" "ERROR" "create_hier_cell_mb3_timers_subsystem() - Empty argument(s)!"}
+     catch {common::send_msg_id "BD_TCL-102" "ERROR" create_hier_cell_mb3_timers_subsystem() - Empty argument(s)!"}
```
yunqu: I think this is a typo here, because the quotes do not match. My original changes on these lines were correct.

schelleg: we discussed offline - this is what Vivado is generating and we'll monitor if issue persists when we upgrade Vivado versions later.

PR #272, opened 2017-02-03
Interrupt Refinements

PeterOgden:
This request adds interrupt support the IOP and updates the display function in LCD class to use asyncio;  fixes an exception when no interrupts are present in the design; and adds a CPU monitor to the asyncio demo.

### DIFF
--- a/python/pynq/iop/iop.py
+++ b/python/pynq/iop/iop.py
@@ -1,30 +1,30 @@
 #   Copyright (c) 2016, Xilinx, Inc.
 #   All rights reserved.
-# 
-#   Redistribution and use in source and binary forms, with or without 
+#
+#   Redistribution and use in source and binary forms, with or without
 #   modification, are permitted provided that the following conditions are met:
 #
-#   1.  Redistributions of source code must retain the above copyright notice, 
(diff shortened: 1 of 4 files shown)


python/pynq/iop/iop.py
```diff
@@ -1,48 +1,85 @@
...
+        intr_ack_gpio : int
+            Number of the GPIO pin used to clear the interrupt
+
+        """
+        self.interrupt = Interrupt(intr_pin)
+        self.gpio = GPIO(GPIO.get_gpio_pin(intr_ack_gpio), "out")
+
+    @asyncio.coroutine
+    def wait(self):
+        """Coroutine to wait until the event is set by an interrupt"
```
yunqu: Need to delete the " at the end of the line.

python/pynq/iop/iop.py
```diff
@@ -1,48 +1,85 @@
...
+
+        """
+        yield from self.interrupt.wait()
+
+    def clear(self):
+        """Clear the interrupt and reset the event
+
```
yunqu: same here. more instructions / docstring?

PR #278, opened 2017-02-22
fix bugs in class my_new_accelerator

sonnyhcl:
data_address should be increased as data goes while read or write data using mmio
fix ```self.mmio.write(RAW_DATA_OFFSET, raw_data[i])``` to ```self.mmio.write(RAW_DATA_OFFSET+4*i, raw_data[i]```

### DIFF
--- a/Pynq-Z1/notebooks/examples/overlay_integration.ipynb
+++ b/Pynq-Z1/notebooks/examples/overlay_integration.ipynb
@@ -167,7 +167,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 24,
+   "execution_count": 1,
    "metadata": {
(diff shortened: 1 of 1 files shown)


yunqu:
Thanks for pointing out this issue. I agree on what you have changed.

For this notebook, can you restart the kernel and rerun all the cells? In that case, the cell count will start from 1, instead of 24.

sonnyhcl:
I have restarted the kernel and reruned the cells. 

Any more comments?

yunqu:
That is fine. Please give us more time to merge your request.

sonnyhcl:
Hi Rock,
Thanks for your polishing. I've also learned a lot from that.
Sonny

PR #280, opened 2017-03-04
Add wifi autoconnect

JorisBolsens:
Check for wifi config script in root of boot partition and attempt to connect if it exists.

this script will mount the boot partition under /boot, and check for "wpa_supplicant.conf" if it exists it will attempt to connect using that as the config file.

### DIFF
--- a/scripts/linux/5_wifi_connect.sh
+++ b/scripts/linux/5_wifi_connect.sh
@@ -5,8 +5,8 @@ set -e
 mount="/boot"
 config_name="wpa_supplicant.conf"
 
-#use fdisk to list devices, fine the one with "*" under Boot and filter out ram
-boot_part="$(fdisk -l | awk '/\*/' | grep -v '512' | cut -d ' ' -f1)"
+#use the 1st partition of sdcard
+boot_part="/dev/mmcblk0p1"
 
 #mount boot partition to temp folder
 mount -o rw "$boot_part" "$mount"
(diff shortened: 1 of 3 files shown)


schelleg:
@JorisBolsens - thanks for the commit.  This is a useful feature for sure.  Give us a day or two to test, comment and pull.

schelleg:
@JorisBolsens - would you rather have the wpa_supplicant.conf on the EXT4 filesystem (e.g.  somewhere like /etc/wpa_supplicant/wpa_supplicant.conf.pynq)?  I want to understand if useful having that file on the FAT boot filesystem - I would expect that file to be generated on PYNQ or do would you expect to generate it from other machines?

That said - I filed issue for next image to include boot partition mounted by default.

JorisBolsens:
@schelleg 
The usefulness of the config being on the boot partition is so that it can be written to the SD card by the user without needing to connect to PYNQ via seriel or ethernet. This makes it so the user can just write the pynq image to the SD card, add their network setup, plug it in and boot it up, which is great because many times ethernet is in an inconvenient location and USB-seriel drivers are notoriously unreliable. It may be better to move it off the boot partition to a standard location as you mentioned, instead of keeping it there, however if it is left on the boot partition it is more easily modified if network conditions change (user can just plug sd card back into pc and modify as necessary). 
Another solution that could accomplish both goals (moving to standard location and easily modified by user on pc) would be to check boot partition for the config and overwrite the existing config in the standard location if a new version exists on the boot partition.

JorisBolsens:
Another helpful change would be if the udev rules were updated to use the wlan# naming scheme instead of wlx[mac address]

schelleg:
got it - I agree for quick wifi readiness when no serial/ethernet available - something I am taking for granted right now.

PR #285, opened 2017-03-20
Augmented MMIO and added python VDMA Driver

cospan:
Added some helper functions to MMIO, these functions simplify interfacing with registers with the following functionality:

-set/clear a bit in a register
-read an individual bit from a register
-read/write a range of bits from a register using both range values and a bitmask

Added a vdma python driver that interfaces with Video DMA core. In order to support this it was necessary to add a framebuffer module.

An example notebook can be found here:

https://github.com/CospanDesign/pynq_notebook_vdma_test/blob/master/notebook/VDMA%20Test.ipynb

A Pynq image can be found here:
https://github.com/CospanDesign/pynq_notebook_vdma_test/tree/master/overlay

There is also a unit test for the vdma in

PYNQ/python/pynq/drivers/test/test_vdma.py

Documentation for the VDMA can be found in the module and notebook.

Dave

### DIFF
--- a/python/pynq/_pynq/_video/_framebuffer.c
+++ b/python/pynq/_pynq/_video/_framebuffer.c
@@ -1,0 +1,321 @@
+/******************************************************************************
+ *  Copyright (c) 2016, Xilinx, Inc.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *
+ *  1.  Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *
+ *  2.  Redistributions in binary form must reproduce the above copyright
+ *      notice, this list of conditions and the following disclaimer in the
+ *      documentation and/or other materials provided with the distribution.
+ *
+ *  3.  Neither the name of the copyright holder nor the names of its
+ *      contributors may be used to endorse or promote products derived from
+ *      this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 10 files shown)


python/pynq/_pynq/_video/_framebuffer.c
```diff
@@ -0,0 +1,321 @@
...
+    unsigned int ret = cma_get_phy_addr(self->frame_buffer);
+    return PyLong_FromUnsignedLong(ret);
+}
+
+/*
+ * get a bytearray of the framebuffer
+ */
+PyObject *get_framebuffer(framebufferObject *self)
+{
+    return PyByteArray_FromStringAndSize((char *)self->frame_buffer, self->width * self->height * self->color_depth);
```
yunqu: Is it possible to break this long line to 2, in order to make sure all lines do not exceed 79 characters?

cospan: Oops, That's what I forgot to do, I'll get that done.

cospan:
Okay, I think I got them all, For some reason my 80 char highlight doesn't work on 'c' files.

schelleg:
hi @cospan - thank you for the contribution.  

We will need  time to integrate the VDMA driver - we are in the process of splitting out the VDMA from the HDMI frontend and then we'd evaluate the driver.

For register class - have you considered bitarrays with slicing?  Slicing is preferable to function calls.

cospan:
@schelleg 
Awesome, I'll incorporate your teams changes when they come in. I appreciate the comment on the bitarray and I understand that there is probably a significant overhead when it comes to function calls. It was just easier to not worry about a register shadow copy at the moment.

Regarding your VDMA tests. I've created a test overlay for two connected VDMAs, it is not that interesting and I only used them for testing the VDMA driver but I know they work and they are independent of HDMI.

[VDMA Ovelay](https://github.com/CospanDesign/pynq-hdl/tree/master/Projects/Simple%20VDMA)

The PDFs are the block diagrams for the Vivado Project so you won't need to open Vivado to view them. I've also included bit and tcl files.

PR #286, opened 2017-03-22
Add helper function to interact with DMA driver using numpy

PeterOgden:
This commits adds a helper function that will create a numpy array from the buffer and optionally resize the buffer if a shape is provided. Also slightly tweaks the transfer function to pick declared defaults for the length and transfer direction.


yunqu:
Hi, I think the changes you made are good. But I am just not sure whether we should directly work on the master branch, or work on the image_v1.5 release branch (helps to keep a clean master.)

PR #288, opened 2017-04-07
pmod_tc1 and tracebuffer fix

yunqu:
Small code polishing / docstring fix for modules:

1. pmod_tc1, with notebook verified.

2. tracebuffer, with the max sample rate corrected.


yunqu:
@cathalmccabe   I don't agree on the pmod_tc1 changes. Please have a look at all of our other codes:
https://github.com/Xilinx/PYNQ/blob/master/Pynq-Z1/sdk/pmod_adc/src/pmod_adc.c#L161
https://github.com/Xilinx/PYNQ/blob/master/Pynq-Z1/sdk/pmod_grove_adc/src/pmod_grove_adc.c#L186

If you think the original one is the convention that we should follow, we should make changes to all of our other codes as well.

cathalmccabe:
@yunqu yes, agreed, there are examples of doing this both ways. Pmod_als and pmod_tmp2 examples of the "original" way. 
Yes, we should make this consistent, so before accepting this pull let's decide and try to update whatever other modules. 

With the original code, there is nothing to stop a "cancel" command being send before issuing the next command so I don't think this is confusing, just slightly more flexible/efficient or simpler to write a driver. Plus, with the acknowledge, some indication the command was received.

yunqu:
Closing for now. Will merge into a later fix.

PR #289, opened 2017-04-11
Notebook 5.0.0 and higher require an additional option when run as root

PeterOgden:
For security reasons jupyter notebook server no longer runs as root
unless an explicit option is provided. This change modifies the boot
script to test for the version and add it if necessary.


schelleg:
This is getting added into image build flow - will close and wait for those pull-requests.

PR #290, opened 2017-04-13
MMIO r/w parameter tests

plysaght:
1.

`or test length < 0
`

 is masked by the test on the previous line

2.

`if offset % 4:`

is simpler than 

```
idx = offset >> 2
if idx << 2 != offset:

```


plysaght:
Good points, I will make the necessary changes.

yunqu:
I have opened another pull request based on suggestions. This one can be closed safely.

PR #291, opened 2017-05-10
Fix for RTD build

cathalmccabe:
overlay tcl and libsds_lib.so not available in RTD build,
and cause doc build to fail

* Add exception to PL to return none if reading tcl fails
* Add exception to Xlnk to check loading libsds_lib.so
* Add check for x86_64 architecture to Xlnk, and skip
  load of libsds_lib.so if true


schelleg:
will reopen against v1.5_image

PR #293, opened 2017-05-19
Video rework

PeterOgden:
This commit changes the video API to make the following changes
 - Separate the drivers for the VDMA and HDMI frontend
 - Add support for hardware colorspace conversion and bits per pixel
 - Give a zero-copy API to reading and writing video frames
 - Asyncio support for reading and writing frames at the refresh rate

There are also two example notebooks included one for the high-level pipeline API for the Pynq-Z1 and one introducing each stage separately.


PeterOgden:
Good catch - merge failure on my part. Do you want to try the updated version?

PR #296, opened 2017-05-26
V1.5 microblaze restructure

schelleg:
Microblaze subsystems in independent folders
PynqMicroblaze superclass for common soft processor methods


schelleg:
@yunqu will re-open with updated pull-request from his branch

PR #307, opened 2017-06-22
Tidy up some PL server details

PeterOgden:
* Ensure the PL server is restarted after an install
 * Use setuptools to generate the start and stop PL server scripts
 * Raise a more obvious exception when connection to the PL server fails


pynq/pl.py
```diff
@@ -536,7 +536,11 @@ def client_request(cls, address=PL_SERVER_FILE,
         None
 
         """
-        cls._remote = Client(address, family='AF_UNIX', authkey=key)
+        try:
+            cls._remote = Client(address, family='AF_UNIX', authkey=key)
+        except FileNotFoundError:
+            raise ConnectionError(
+                       "Could not connect to Pynq PL server") from None
```
yunqu: @PeterOgden I think we can ignore the types of error; that is, no matter what exception occurs, we raise the ConnectionError / RuntimeError.

pynq/pl.py
```diff
@@ -683,6 +687,19 @@ def __init__(self):
...
 
 
+def _stop_server():
+    try:
+        PL.client_request()
+        PL.server_update(0)
+    except:
+        pass
+
+def _start_server():
```
yunqu: @PeterOgden   Can we add 2 or 3 sentences as docstrings so users know what those 2 functions are doing?

pynq/pl.py
PeterOgden: My worry with being more general in the except block is that there may be other circumstances where a different error may be thrown which may be caused by things other than PL server not being started. For example some form of serialiasation error caused by the PL server and the process running different versions of python. In these instances I think we would want to pass the full exception back up the stack. The aim for this commit is to very specifically change a FileNotFoundError on a file that the user has never seen and present something more useful.

PR #308, opened 2017-06-26
Fix ReadTheDocs build errors

drichmond:
Fix ReadTheDocs build errors

* Added a standard architecture-check in ps.py
* Added build guards to prevent exceptions when executing on RTD virtual machines (x86_64)
* Added warnings when guards detect an unsupported architecture using the aptly-named warnings package
* Added Notes in doc-strings to alert users of abnormal behavior when unsupported architectures are used
* Changed python3.6 style format strings to python3.5-style format strings
* Corrected a path for the new repository structure
* Refactored guards in xlnx.py 
* Moved cffi/libffi calls into xlnx.py class scope from global scope and added guards
* Moved cffi/libxlnk/libmemapi into dma.py class scope from global scope and added guards

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -19,7 +19,7 @@ import os
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('../../python'))
+sys.path.insert(0, os.path.abspath('../../'))
 
 import pynq
 
--- a/pynq/interrupt.py
+++ b/pynq/interrupt.py
@@ -59,7 +59,7 @@ class Interrupt(object):
 
         """
         if pinname not in PL.interrupt_pins:
-            raise ValueError("No Pin of name {0} found".format(pinname))
+            raise ValueError("No Pin of name {} found".format(pinname))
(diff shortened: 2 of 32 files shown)


drichmond:
I'd be happy to remove the warnings. I added them because I found some of them useful in @cathalmccabe's branch. I don't believe they will any additional runtime to the typical (read, Pynq) use case.

yunqu:
I guess the reason why we downgraded 3.6 to 3.5 is readthedocs does not support 3.6? This seems tricky since we cannot always do this (there are too many places using 3.6).

drichmond:
We hope that in the long run RTD upgrades to 3.6, but in the short term we can't use 3.6 format strings. 

Fortunately the number of spots that use 3.6 is fairly limited

cathalmccabe:
Yes, RTD is 3.5.
We can manage this by downgrading formatted strings.

Cathal

On 27 Jun 2017, at 17:52, Yun Rock Qu <notifications@github.com<mailto:notifications@github.com>> wrote:


I guess the reason why we downgraded 3.6 to 3.5 is readthedocs does not support 3.6? This seems tricky since we cannot always do this (there are too many places using 3.6).

drichmond:
RTD has 3.6 support in their beta. Instead of downgrading, we could try the beta version?

(I'll submit a request anyway)

PR #309, opened 2017-06-27
 Rework overlay infrastructure to aid discoverability

PeterOgden:
This commit introduces a set of changes to the Overlay class to expose hierarchies and IP as attributes on the instance.

 * Introduce a set of Default implementations for IP, hierarchy and overlay drivers
 * Search for custom Overlay classes when loading an overlay
 * Register subclasses of DefaultIP and DefaultHierarchy so drivers can be reused.
 * Separate out Overlay into a separate file to reduce circular dependencies.
 * Add the set of pins to the gpio_dict so they can be exposed.
 * Update existing drivers to be automatically registered
 * Add new AXI GPIO and AXI DMA drivers
 * Update base overlay with custom python class
 * Update video, starting and Pmod OLED notebooks.


drichmond:
At risk of opening a can of worms - but  can we in theory agree on a method naming scheme? I like the re-work, but we have methods like load_ip_dict and find_entry, mixed with methods like builddocstring and completedescription in the same file (pl.py)

As an example: it seems that we tacitly agreed that classes are CamelCase. 

I realize that this heavily dependent on personal coding style

yunqu:
@drichmond  Agreed. Based on https://www.python.org/dev/peps/pep-0008/#function-names we should use underscore to improve readability.

PeterOgden:
Agreed on the inconsistency. I read the PEP as saying underscores should be used "as necessary" to improve readability and tend to omit them. Python itself is pretty inconsistent on this with `weakref.getweakrefcount` and `asyncio.get_event_loop`. I'm willing to go with majority opinion on this.

tghastly:
Guys, pls use the PEP 8 guidelines.  Peter's comment is correct but as PEP 8 notes 

"The naming conventions of Python's library are a bit of a mess, so we'll never get this completely consistent -- nevertheless, here are the currently recommended naming standards. New modules and packages (including third party frameworks) should be written to these standards, ..."

That said, and esp wrt the 79 character guidance, the better approach, as cited elsewhere, is ...
"As George Orwell said about his own guidelines: break any of them sooner than doing anything outright barbarous.

This is a useful ref 
Raymond Hettinger - Beyond PEP 8 -- Best practices for beautiful intelligible code - PyCon 2015

So CamelCase for classes and u_scores for variables, etc

PeterOgden:
I've added some underscores to separate out long names and some other tidying,

PR #310, opened 2017-06-27
improve mmio readability and small bug fix

yunqu:
This is a small fix for the mmio class.

### DIFF
--- a/pynq/mmio.py
+++ b/pynq/mmio.py
@@ -27,14 +27,13 @@
 #   OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
 #   ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-__author__ = "Yun Rock Qu"
-__copyright__ = "Copyright 2016, Xilinx"
-__email__ = "pynq_support@xilinx.com"
-
-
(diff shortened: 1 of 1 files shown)


pynq/mmio.py
```diff
@@ -128,18 +127,17 @@ def read(self, offset=0, length=4):
...
-        if not length == 4:
+        if length != 4:
             raise ValueError("MMIO currently only supports 4-byte reads.")
-        if offset < 0 or length < 0:
-            raise ValueError("Negative offset or negative length.")
-        idx = offset >> 2
-        if idx << 2 != offset:
-            raise MemoryError('Read operation unaligned.')
+        if offset < 0:
+            raise ValueError("Negative offset.")
```
cathalmccabe: Small point, could the error message be made more helpful? It is currently stating what the offset is, rather than what you should do to correct it. 
read(-4):
i.e. Replace with "Offset cannot be a negative value"?

Also, the code around ~line 77 should probably be modified in the same way as read/write checks to seperate the check for base_addr, and length. 

```
if base_addr < 0 or length < 0:
    raise ValueError("Negative offset or negative length.")
```
Also, the error message is not correct. This is referring to the base address, not offset. Could this be corrected in this commit from offset -> base address?

PR #318, opened 2017-07-07
V1.5 DIO API change

yunqu:
This pull request includes the following change:

1. Python API has changed for all the builders. Builders now have to follow a flow similar to: init() --> setup() --> run() --> stop().

2. Dummy state has been added to FSM machine so users will see consistent initial state in the 1st sample of trace analyzer. Dummy sample is also added to pattern builder to make sure FSM and pattern builders work in sync.

3. Boolean builder bugs have been fixed so in the RUNNING state, logic outputs are held on output pins.

For now this branch should work but remaining changes are expected for this pull request soon:

* Will update names in SDK project and tcl file so names like SMG, PG, CFG will disappear.

* Will update Python constants based on the above name change.


yunqu:
@drichmond  Your changes on these files make my pull request conflicting... This produces more work for me, so next time please leave these files untouched. I can do the formatted string by myself.

drichmond:
I was told to fix all of the format strings by yesterday afternoon - sorry if it caused any conflicts, but those were my instructions.

PR #320, opened 2017-07-11
Generate the Trace Controller IP

PeterOgden:
Use HLS to generate the IP rather than include the generated sources in the repository


drichmond:
Perhaps this is a meta-question: is there any reason we shouldn't, or can't, have the HLS IP be a makefile rule, with its own makefile that is called from the overlay makefile?

PR #363, opened 2017-07-25
updated waveform to be able to use phantomjs

yunqu:
2 ways are implemented to render wavedrom : javascripts, or phantomJS.


pynq/lib/logictools/waveform.py
```diff
@@ -176,24 +173,188 @@ def draw_wavedrom(data):
...
-    if not os.path.isfile(PYNQ_JUPYTER_NOTEBOOKS + wavedrom_js):
-        raise RuntimeError('Cannot locate {}.'.format(wavedrom_js))
-    if not os.path.isfile(PYNQ_JUPYTER_NOTEBOOKS + wavedromskin_js):
-        raise RuntimeError('Cannot locate {}.'.format(wavedromskin_js))
+    wavedrom_js = 'wavedrom.js'
+    wavedromskin_js = 'wavedromskin.js'
+
+    if not (_is_javascript_present(wavedrom_js) and
+            _is_javascript_present(wavedromskin_js)):
+        _copy_javascripts()
```
PeterOgden: Why are we copying the Javascript? Why don't we just do what nbwavedrom does and insert the wavedrom code into the notebook output?

pynq/lib/logictools/waveform.py
```diff
@@ -176,24 +173,188 @@ def draw_wavedrom(data):
...
+
+    This method is only required when rendering the wavedrom using 
+    javascripts. This is not required for PhantomJS.
+
+    """
+    src_folder = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), 'js')
+    dst_folder = PYNQ_JUPYTER_NOTEBOOKS
+    if os.system('cp -rf ' + src_folder + ' ' + dst_folder):
+        raise RuntimeError('Cannot copy the javascripts.')
```
PeterOgden: If we are going to copy what's wrong with `shutil.copytree`? Forking a shell seems somewhat like overkill

pynq/lib/logictools/waveform.py
yunqu: 1st reason is I am not sure whether phantomjs is going to be available for future boards, 2nd reason is that the nbwavedrom does not do a good job rendering long waveform, so perhaps we will fall back to the original version...

PR #372, opened 2017-07-27
Changes to discovery based on feedback

PeterOgden:


drichmond:
Are we getting rid of the Overlay = <OverlayName> at the bottom of the python files?

PR #452, opened 2017-08-18
V2.0 misc docs

yunqu:


docs/source/faqs.rst
```diff
@@ -180,23 +182,24 @@ board's static IP here.
...
-      /etc/hostname
+host names.  You can change the hostname by editing the Linux hostname files:
+``/etc/hostname`` and ``/etc/hosts``.
    
 What is the user account and password?
 --------------------------------------
 
-Username and password for all Linux, jupyter and samba logins are:
-``xilinx/xilinx``
+Username and password for all Linux, Jupyter and Samba logins are:
```
drichmond: I split this in Getting started - can you split into somethin like : The username for Jupyter and Samba is **xilinx** and the password is **xilinx**?

Just in case someone gets confused and thinks the username is xilinx/xilinx

PR #466, opened 2017-10-13
Corrected errors in MMIO code example by replacing it with latest ver…

plysaght:
Corrected errors in MMIO code example by replacing it with latest version from workshop notes

got rid of errors due to using self.mmio. ...
changed rest to match example in workshop notes - data is now 0x12345678
updated range top addr to 0x40000400


schelleg:
pulling in #489 which has same docs/ content but onto image_v2.1 branch.

PR #470, opened 2017-10-31
fix arduino_io.py

yunqu:


yunqu:
I will close it since it will be fixed in my other working branch.

PR #477, opened 2017-11-08
fix usb_wifi to handle multiple interfaces

yunqu:


pynq/lib/usb_wifi.py
```diff
@@ -51,7 +52,7 @@ class Usb_Wifi(object):
 
     """
 
-    def __init__(self):
+    def __init__(self, interface=None):
```
schelleg: interface='wlan0'

we will be using this naming in 2.1

pynq/lib/usb_wifi.py
```diff
@@ -64,8 +65,12 @@ def __init__(self):
         for line in net_devices.splitlines():
             m = re.match('^([\d]): ([\w]+): *', line)
             if m:
-                if m.group(2) != 'lo' and m.group(2) != 'eth0':
+                if interface is None and m.group(2).startswith('wl'):
+                    self.wifi_port = m.group(2)
+                    break
+                elif interface is not None and m.group(2) == interface:
                     self.wifi_port = m.group(2)
+                    break
```
schelleg: this can then be simplified and perhaps use something more pythonic to find interface list?

netifaces has simple calls to get list of interfaces - maybe that or something else

Issue #490, opened 2017-12-20
Small tcl parsing bug in pl.py


a1650078:
While creating a minimal example for faster overlay testing, I tried to save time by using vivado's save/apply configuration function under the presets menu in ip configuration. This worked fine, but seems to add lines to the tcl for the block diagram. They look like this

`CONFIG.PCW_FCLK2_PERIPHERAL_DIVISOR1.VALUE_SRC {DEFAULT} \`

and the regex `clk_odiv_regex` picks them up and confuses the parser. Changing the line that tests the ignore regex to also skip if `"VALUE_SRC" in line` fixes it.

yunqu:
Should be fixed in https://github.com/Xilinx/PYNQ/pull/491

PR #491, opened 2017-12-20
fix pl.py to handle more corner cases

yunqu:
1. add clock parsing for zynq ultrascale devices

2. fix parsing when PL clock `enable` is not set explicitly

3. skip parsing the `{DEFAULT}` configurations

Should fix following issues:
https://github.com/Xilinx/PYNQ-Dev/issues/97
https://github.com/Xilinx/PYNQ/issues/490

### DIFF
--- a/pynq/pl.py
+++ b/pynq/pl.py
@@ -159,6 +159,7 @@ class _TCL:
         self.ip_dict = {}
         self.gpio_dict = {}
         self.clock_dict = {}
+        self.family = "xc7z"
 
         # Key strings to search for in the TCL file
         family_pat = "create_project"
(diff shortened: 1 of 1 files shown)


a1650078:
I'm concerned that this won't fix all future instances of #490 because the value can be `{DEFAULT}` but also `{USER}`. I think you should ignore all `VALUE_SRC` directives instead but I have no idea if that is sensible or not.

yunqu:
@a1650078 Let me know if this commit 366016e is okay.

a1650078:
Yep, this fixes the issue I had when using save/apply presets in vivado. Thanks!

PR #500, opened 2018-01-14
libsds update

yunqu:
Updated to 2017.4 version.


PeterOgden:
We should probably do this at the same time as we update the kernel version in the boot config otherwise the image may break.

yunqu:
Thanks Peter. Will add soon.

yunqu:
@PeterOgden  Let me know if this is okay.

PR #515, opened 2018-01-26
LegacyDMA fix

yunqu:


yunqu:
closing this for now because this will break setup.py. Need to look into this further.

PR #541, opened 2018-02-09
purge opencv 2.4.9 since opencv3.1 is present

yunqu:
removing opencv2.4.9 to be able to pass verification test


schelleg:
next time, anything we can do to break up the massive package list?  Makes history unreadable.

PR #563, opened 2018-02-13
Add IPython Microblaze Documentation

PeterOgden:


docs/source/pynq_libraries/pynq_microblaze_subsystem.rst
```diff
@@ -64,4 +64,36 @@ More information about the PynqMicroblaze class, and its API can be found in the
...
+PYNQ provides a TCL script to generate the BSP from the hardware description
+file which can be found in the ``boards/sw_repo`` directory of the repository,
+along with the drivers needed for Python/C communication and the ``pynqmb``
+hardware abstraction library. Creating and using the BSP requires the following
+steps:
+
+1. Export Hardware from Vivado to generate the HDF file
+
+2. In the ``boards/sw_repo`` directory run ``xsct build_xsdk.tcl $HDF_FILE hw_project``
+
```
yunqu: I think we can run `make` here since we have a makefile (looks better than the `xsct` call?)

PeterOgden: `make` will only recreate the BSPs for the base overlay though. I wanted to explain how people could create BSPs for their own custom instances. I accept that we are missing the part about running `make` to create the base overlay BSPs though.

docs/source/pynq_libraries/pynqmb_reference.rst
```diff
@@ -0,0 +1,252 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Closes a SPI master
+
+Timer Devices
+-------------
+
+Timer devices serve two purposes. They can either be used to output PWM
+signals or as program timers for inserting accurate delays. All of these
+functions are in ``timer.h``
```
yunqu: should we also mention that if a timer is used for `delay_ms` and such, it cannot be used for other functionalities?

docs/source/pynq_libraries/pynq_microblaze_subsystem.rst
PeterOgden: Or we can call `make HDF=$HDF` which I agree looks better

docs/source/pynq_libraries/pynqmb_reference.rst
PeterOgden: Yes we should

docs/source/pynq_libraries/pynqmb_reference.rst
```diff
@@ -0,0 +1,252 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Issue a write command to the specified slave.
+
+``void i2c_close(i2c dev_id)``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Close the I2C device.
+
+SPI Devices
```
yunqu: SPI Devices
-------------
To be consistent with other titles.

PR #623, opened 2018-05-04
add hackathon drivers using ipython microblaze

yunqu:


yunqu:
will reopen against image_v2.3 branch.

PR #646, opened 2018-07-06
Replace self with Xlnk

davidbrochart:


schelleg:
Hi @davidbrochart -  thank you for looking at that.

@sfox14 pulled this onto our image_v2.3 branch here: https://github.com/Xilinx/PYNQ/pull/640 ...  great minds think alike.
 
We're planning on merging onto master shortly.

PR #656, opened 2018-07-16
Update pynq_sd_card.rst

ly0:
This PR is fullfilled my complaints about so many problems in this flow and finally I reached the end of this marathon.

IT REALLY TAKES ME HOURS.


schelleg:
We'll be improving the sdbuild flow in the next release.

PR #659, opened 2018-07-17
wpa_ap: Accept some delay in wlan0 coming up

PeterOgden:


sdbuild/packages/wpa_ap/wpa_ap/ap.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 
 # User can adjust wlan0 to connect to a specific wifi network
-ifconfig wlan0 up
+while ! ifconfig wlan0 up
```
yunqu: Should we add some constraint condition to prevent it hanging?

PeterOgden: systemd will kill it off after a couple of minutes. We could add an explicit timeout here though if you prefer

PR #679, opened 2018-08-14
pynqmb: SPI transfer supports read w/o write (and visa versa)

jackrosenthal:
This is changes that we made to use pynqmb with PyBlaze, the SPI transfer function should support the ability to write without reading, and the ability to read without writing. The simple format to do this was to simply pass ``NULL`` for either ``read_data`` or ``write_data``.

We also cleaned up some sloppy/inconsistent coding style in this module.

### DIFF
--- a/boards/sw_repo/pynqmb/src/spi.c
+++ b/boards/sw_repo/pynqmb/src/spi.c
@@ -174,13 +174,20 @@ void spi_transfer(spi dev_id, const char *write_data, char *read_data,
     unsigned volatile char j;
     unsigned int base_address = xspi[dev_id].BaseAddr;
 
+    XSpi_WriteReg(base_address, XSP_SSR_OFFSET, 0xfe);
     if (write_data) {
-        XSpi_WriteReg(base_address, XSP_SSR_OFFSET, 0xfe);
         for (i = 0; i < length; i++) {
             XSpi_WriteReg(base_address, XSP_DTR_OFFSET, write_data[i]);
         }
-        while (((XSpi_ReadReg(base_address, XSP_SR_OFFSET) & 0x04)) != 0x04);
     }
+    else {
+        /* Behavior for no data to write while reading is to write
+         * zeroes */
+        for (i = 0; i < length; i++) {
+            XSpi_WriteReg(base_address, XSP_DRR_OFFSET, 0);
+        }
+    }
+    while (((XSpi_ReadReg(base_address, XSP_SR_OFFSET) & 0x04)) != 0x04);
 
     // delay for 10 clock cycles
     j = 10;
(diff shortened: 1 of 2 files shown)


PeterOgden:
Hi Jack

Thanks for the contribution and I agree with the changes. The one thing I'm unsure of is whether it is valid to not read/write to the data registers. I've been looking through the data sheet and it seems to imply that the registers should always be written/read from even if the data is either dummy write data or ignored read data. Do you have any knowledge/experience of this? We can always add dummy writes in if `write_data` is `NULL` and not store anything in `read_data`.

Peter

jackrosenthal:
Hey Peter,

Are you looking at PG153? p. 100 makes it seem like write and read can be issued individually.

One other thing I think: I think the write to SSR and check on SR for write can go into the if, as well as the delay, etc. for the read. I will test this change and commit if it works.

Jack

jackrosenthal:
OK, so after a bit of testing on an oscilloscope here is what I determined:

* Yes, the SPI core seems to behave normally with write-only or read-only *with the initial code I committed*. When the delay is removed from write-only operation, some of the bytes get truncated after the first call to ``spi_transfer`` for write-only operations.
* It is OK to remove the initial steps for read-only operation (see latest commit).
* Read-only operation does not really make sense: it would essentially be equivalent to constantly writing zeroes while reading... this is how it behaved when I tested it passing ``NULL``.

So I think what is there now should be working when ``NULL`` is passed for either parameter.

cathalmccabe:
I think the Data Transmit Register and Data Receive Register can be registers or FIFOs depending on the core configuration. If FIFOs, I would have though the dummy bytes are needed for the transfer.
We just need to be careful all uses cases are tested before accepting, and that this is not an optimization for the particular configuration of the AXI SPI we have used in the "base", or other example designs.

jackrosenthal:
I would agree, it is probably worth writing dummy bytes even when ``read_data`` is ``NULL``, this way we have a well defined behavior regardless of configuration of the SPI core.

I will add this to the commits.

Jack

jackrosenthal:
OK, so with the latest commit, here is the defined behavior:

* When ``write_data`` is ``NULL``, zeroes will be written equivalent to ``length``
* When ``read_data`` is ``NULL``, the data will not be read into any buffer

jackrosenthal:
Hey guys,

When do you think this can get merged?

Also, should this be merged into master or image-v2.3 branch?

Jack

yunqu:
@jackrosenthal 
Thanks for your contribution on PYNQ. We really appreciate this effort.

The problem with this pull request is that there are a lot of formatting changes in the code, making it hard for us to review. Can you keep the code change as minimum, as most of the code changes do not seem to be necessary?

Another thing is that the bsp's have to be recompiled in order to reflect your changes. I can cherry-pick your commit, along with other fixes/improvements we have on our side, and update other necessary files.

PR #688, opened 2018-08-21
Kevja branch 2

kevja:
Updated led, buttton, and switch classes to use axigpio class with ABCs defined in gpio_abc.py in 'lib' folder


schelleg:
will reopen later ...

PR #695, opened 2018-08-22
add zcu104 bsp generation

yunqu:


boards/ZCU104/petalinux_bsp/meta-user/recipes-kernel/linux/linux-xlnx/zcu104_ina226.patch
```diff
@@ -0,0 +1,27 @@
+diff -u b/arch/arm64/boot/dts/xilinx/zynqmp-zcu104-revA.dts b/arch/arm64/boot/dts/xilinx/zynqmp-zcu104-revA.dts
```
PeterOgden: I don't believe this patch is necessary any more - what it did is better done in the device-tree patch.

PR #716, opened 2018-09-05
Use FPGA manager for Zynq-7000 designs

PeterOgden:
Looks like xdevcfg has disappeared but luckily the ZU+ code download works

### DIFF
--- a/pynq/pl.py
+++ b/pynq/pl.py
@@ -1602,7 +1602,6 @@ class _BitstreamMeta:
 
         """
         self._download()
-        self._update_pl()
         if not self.partial:
             self._update_pl()
 
--- a/pynq/pl.py
+++ b/pynq/pl.py
@@ -1623,53 +1623,8 @@ class _BitstreamMeta:
         PL.server_update()
 
 
-class _BitstreamZynq(_BitstreamMeta):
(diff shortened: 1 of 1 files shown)


parimalp:
Hi Peter,

You already might know that configuration in MPSoC is handled by CSU unit. There is also an FPGA configuration library, at least for standalone OS, that provides simple API to transfer bitstream.

Regards,

Parimal

pynq/pl.py
```diff
@@ -1623,52 +1623,7 @@ def _update_pl(self):
...
-        PL.shutdown()
-
-        # Write bitfile to xdevcfg device
-        with open(self.BS_XDEVCFG, 'wb') as f:
-            f.write(buf)
-
-
-class _BitstreamUltrascale(_BitstreamMeta):
+class Bitstream(_BitstreamMeta):
     """This class instantiates a PL bitstream for Zynq Ultrascale.
```
yunqu: one small comment: maybe update the docstring?

yunqu:
i guess this PR will conflict with #715. Both of these 2 PRs modify pl.py. Need to fix.

PeterOgden:
Doesn't conflict - I've tried the merge

PR #725, opened 2018-09-07
Update ZCU104 to make sure the MAC address is read from the EEPROM

PeterOgden:


boards/ZCU104/petalinux_bsp/meta-user/recipes-bsp/u-boot/files/platform-top.h
```diff
@@ -0,0 +1,39 @@
+
```
yunqu: Is it sufficient to add `#include <configs/xilinx_zynqmp.h>` to avoid adding our versions of DFU_ALT_INFO_RAM and DFU_ALT_INFO_MMC?

PeterOgden: That file is the one already in the BSP with the addition of the EEPROM lines so I'm assuming it works better than the generic one?

PR #727, opened 2018-09-07
add x11 package for zcu104

yunqu:


yunqu:
reasonable. i will close.

PR #736, opened 2018-09-19
pynqmicroblaze: Perform unsigned conversion correctly and handle hierarchies

stephenneuendorffer:
This commit performs the following fixes
 * pyprintf correctly handles unsigned numbers
 * Correctly handle microblaze hierarchies not at top level
 * Timeout waiting for AXI shutdown managers to close the connection.

### DIFF
--- a/pynq/lib/pynqmicroblaze/rpc.py
+++ b/pynq/lib/pynqmicroblaze/rpc.py
@@ -627,8 +627,11 @@ def _pyprintf(stream):
     args = []
     for i in range(len(format_string)):
         if in_special:
-            if format_string[i:i+1] in [b'd', b'x', b'X', b'o', b'u']:
+            if format_string[i:i+1] in [b'd']:
                 args.append(stream.read_int32())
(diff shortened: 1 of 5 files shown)


pynq/lib/pynqmicroblaze/rpc.py
```diff
@@ -627,8 +627,11 @@ def _pyprintf(stream):
     args = []
     for i in range(len(format_string)):
         if in_special:
-            if format_string[i:i+1] in [b'd', b'x', b'X', b'o', b'u']:
+            if format_string[i:i+1] in [b'd']:
                 args.append(stream.read_int32())
+            elif format_string[i:i+1] in [b'x', b'X', b'o', b'u']:
+                # perform unsigned conversion
+                args.append(stream.read_int32() & 0xffffffff)
```
PeterOgden: We have `stream.read_uint32()` which should do the same thing?

stephenneuendorffer: Sounds like it.

PR #752, opened 2018-11-07
Switch to using c++ for pynqmicroblaze

stephenneuendorffer:
There are several bits to this:
1) define WRAP to make your wrapper generation work. The expectation is that user code has something like:

CODE> %%microblaze overlay.user_port.control_microblaze
CODE> // Must be valid C and C++
CODE> #include <pyprintf.h>
CODE> int foo(int x); // Generate a wrapper for foo()
CODE>
CODE> #ifndef WRAP
CODE> // Must be valid C++
CODE> extern "C" {
CODE>   #include "mb_interface.h"
CODE> }
CODE> int foo(int x) {
CODE>    return x+1;
CODE> }

2) use mb-g++ to compile, with some extra flags to ensure that we don't bring in lots of standard library functions that aren't called
3) Because we have to be careful about code size, dump the size of the largest symbols
4) fix some of the library code to be C++ safe (const correct, declaration correct)

### DIFF
--- a/pynq/lib/pynqmicroblaze/bsp.py
+++ b/pynq/lib/pynqmicroblaze/bsp.py
@@ -85,10 +85,30 @@ class BSPInstance:
         self.libraries = ['xil']
 
         self.linker_script = path.join(root, 'lscript.ld')
-        self.cflags = [
-            '-Os', '-lxil', '-mlittle-endian',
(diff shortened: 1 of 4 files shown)


stephenneuendorffer:
I've added another patch to get the right microblaze options from the BSP.

PR #758, opened 2018-11-14
Tcl refactor

yunqu:


cathalmccabe:
Looks like the make files should have an option to only build the block diagram. i.e. run $(overlay_name).tcl but not build_bitstream.tcl

yunqu:
@cathalmccabe It should be fixed now.

PR #770, opened 2018-12-07
Update audio_playback.ipynb

mr-yamraj:
It is saving in that code block and not playing


yunqu:
Hi,

Thanks for noticing this. I cannot open your notebook in github; the original notebook seems fine to open though:
https://github.com/Xilinx/PYNQ/blob/master/boards/Pynq-Z1/base/notebooks/audio/audio_playback.ipynb

so can we just change the title without deleting cell output?

PR #772, opened 2018-12-10
Fix TCL issue when a hierarchy has multiple sub-hierarchies

PeterOgden:


yunqu:
#772 and #782 better to be merged rather than rebased to avoid merge conflict.

PR #791, opened 2019-01-22
isolate usbgadget from wifi package

yunqu:


PeterOgden:
We should be able to get the same effect by adding isc-dhcp-server.service to the Before clause in service file. It's probably nicer to use systemd to handle dhcpd so we don't end up with it running twice?

PR #802, opened 2019-02-01
Add RegisterMap class and incorporate it into DefaultIP

PeterOgden:
This commit also moves the existing Register class from ps.py to
the new register_map.py and allows the buffer to be provided as
an init parameter. It also adds a simple notebook documenting
the new feature


boards/Pynq-Z1/base/notebooks/register_map_intro.ipynb
```diff
@@ -0,0 +1,332 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Introduction to the Register Map\n",
+    "\n",
+    "In PYNQ 2.4 it is now possible to access the register map for IP in the overlay directly. This functionality requires that the `.hwh` file is distributed along with the bitstream. This can be generated from Vivado.\n",
```
yunqu: I think we can de-emphasize the version number since we will have to update this again in next release.

yunqu:
It'd be better if we could put this notebook inside `notebooks/board` folder? Also, maybe we want to make a symlink for this file in Pynq-Z2/ZCU104 notebooks as well.

PeterOgden:
I've addressed both of those comments - I also noticed that we didn't have the boards folder at all on the ZCU104 because the RGB example doesn't work. The asyncio one does however so I've symlinked that in as well.

PR #808, opened 2019-02-12
fix dma transfer size problem

yunqu:


pynq/lib/dma.py
```diff
@@ -28,20 +28,17 @@
...
 import warnings
-from pynq.ps import CPU_ARCH_IS_SUPPORTED, CPU_ARCH
 
 
 __author__ = 'Peter Ogden, Anurag Dubey'
 __copyright__ = 'Copyright 2017, Xilinx'
 __email__ = 'pynq_support@xilinx.com'
 
 
+DEFAULT_C_SG_LENGTH_WIDTH = 14
```
PeterOgden: My one worry is that this will break any designs out there that don't have a .hwh file and have a length register set > 14 bits. I feel we should throw a warning but not an exception if an over-length buffer is used and no description is available.

PR #809, opened 2019-02-12
clean up pynqmicroblaze package

yunqu:


pynq/lib/pynqmicroblaze/streams.py
```diff
@@ -86,7 +87,7 @@ def read_upto(self, n=-1):
         available = (written - read + self.length) % self.length
         if available == 0:
             return b''
-        if n > 0 and available > n:
+        if available > n > 0:
```
PeterOgden: I'm not sure about this change as the two clauses of the if statement are really testing for different things - `n > 0` is to test if the special cases of `n == 0` or `n == 1` have been passed and that the entire buffer should be returned, while the `available > n` acts a clamp when `n` has been provided. While the two forms are technically equivalent the new one makes that relationship far less obvious and harder to read.

yunqu: fixed.

PR #852, opened 2019-04-19
Sdbuild updates - tool upgrades and dtc

yunqu:


sdbuild/packages/gcc-mb/patches/gcc/5.2.0/0002_microblaze_constant.patch
```diff
@@ -1,25 +0,0 @@
-diff -ur a/gcc/config/microblaze/microblaze.c b/gcc/config/microblaze/microblaze.c
```
PeterOgden: I suspect that we will still need a version of this patch - https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65649

sdbuild/scripts/setup_host.sh
```diff
@@ -53,21 +67,21 @@ fi
 mkdir tools
 cd tools/
 
-wget http://crosstool-ng.org/download/crosstool-ng/crosstool-ng-1.22.0.tar.bz2
-tar -xf crosstool-ng-1.22.0.tar.bz2
-cd crosstool-ng/
+ctver="1.24.0-rc3"
```
PeterOgden: 1.24 is just released

sdbuild/packages/gcc-mb/patches/gcc/5.2.0/0002_microblaze_constant.patch
yunqu: Is there a test code that I can use to check?

PeterOgden: Any code that uses floats on armv7 - I used the following line

     float main() { return 1.4; }

and it's throwing an error

yunqu: Is the error related to the name `main()`? I tried my float operations (this notebook, cell 4: https://github.com/Xilinx/PYNQ/blob/master/boards/Pynq-Z1/base/notebooks/microblaze/microblaze_programming.ipynb)
It looks all good. 

For you example, if I change `main()` to `x()`, it looks good.

Anyway, if you think this patch is necessary,  I will add it back in my next fix.

PeterOgden: If you change the 4th cell in that notebook to 

```C
    float arg_passing(float a, char b, unsigned int c) {
        return a + b + c + 10.0f;
    }
```

I get the same error - 

```
    Compile FAILED
    /tmp/ccZRZv52.s: Assembler messages:
    /tmp/ccZRZv52.s:112: Fatal error: operand must be a constant or a label
```

PeterOgden: Sorry - hadn't noticed the new patch. For some reason I ended up building a 7.4.0 GCC and that missed it.

PR #857, opened 2019-05-16
Update gpio.py to support generic gpio devices

gaskint:
These changes are for the pynq wrapper around Linux's GPIO Sysfs Interface. They make the class more robust in being able to deal with GPIO pins being added and removed from the Linux kernel at run time, such as through device tree overlay's.

### DIFF
--- a/pynq/gpio.py
+++ b/pynq/gpio.py
@@ -39,7 +39,8 @@ __email__ = "pynq_support@xilinx.com"
 class _GPIO:
     """Internal Helper class to wrap Linux's GPIO Sysfs API.
 
-    This GPIO class does not handle PL I/O.
+    This GPIO class does not handle PL I/O without the use of
+    device tree overlays.
 
     Attributes
     ----------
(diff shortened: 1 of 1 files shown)


pynq/gpio.py
```diff
@@ -255,12 +330,49 @@ def get_gpio_pin(gpio_user_index):
...
+
+        This is a static method. To use:
+
+        >>> from pynq import GPIO
+
+        >>> gpio = GPIO.get_gpio_npins()
+
+        Parameters
+        ----------
+        target_label : string
```
yunqu: str

pynq/gpio.py
```diff
@@ -255,12 +330,49 @@ def get_gpio_pin(gpio_user_index):
         ----------
         gpio_user_index : int
             The index specified by users, starting from 0.
+        target_label : string
```
yunqu: str

pynq/gpio.py
```diff
@@ -223,22 +296,24 @@ def get_gpio_base():
         ----
         For path '/sys/class/gpio/gpiochip138/', this method returns 138.
 
+        Parameters
+        ----------
+        target_label : string
```
yunqu: str

pynq/gpio.py
```diff
@@ -209,8 +232,58 @@ def write(self, value):
...
+
+        This is a static method. To use:
+
+        >>> from pynq import GPIO
+
+        >>> gpio = GPIO.get_gpio_base_path()
+
+        Parameters
+        ----------
+        target_label : string
```
yunqu: str

pynq/gpio.py
```diff
@@ -209,8 +232,58 @@ def write(self, value):
...
+            The label of the GPIO driver to look for, as defined in a
+            device tree entry.
+
+        Returns
+        -------
+        string
```
yunqu: str

PR #890, opened 2019-08-13
update to 2019.1 flow and add xrt

yunqu:


sdbuild/boot/meta-pynq/recipes-xrt/xrt/xrt_git.bb
```diff
@@ -0,0 +1,30 @@
...
+                    file://runtime_src/driver/xclng/drm/xocl/LICENSE;md5=b234ee4d69f5fce4486a80fdaf4a4263 \
+                    file://runtime_src/driver/xclng/xrt/user_gem/LICENSE;md5=3b83ef96387f14655fc854ddc3c6bd57 \
+                    file://runtime_src/driver/xclng/tools/xbutil/LICENSE;md5=d273d63619c9aeaf15cdaf76422c4f87"
+
+SRC_URI = "git://github.com/Xilinx/XRT.git;protocol=https;branch=2019.1"
+
+PV = "2.2.0+git${SRCPV}"
+
+# Use latest version
+SRCREV = "${AUTOREV}"
```
PeterOgden: We probably want to lock these for the release once we have a version that works

sdbuild/packages/xrt/xrt_init.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
```
PeterOgden: Does this need to be a service?

sdbuild/boot/meta-pynq/recipes-xrt/xrt/xrt_git.bb
yunqu: I will remove this from meta-pynq layer since the recipes are already inside meta-xilinx-bsp layer  in 2019.1 petalinux.

sdbuild/packages/xrt/xrt_init.sh
yunqu: Will remove this.

sdbuild/packages/xrt/pre.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
giunatale: Apparently this file is causing failure on ZCU104 build. It seems to have wrong permissions (missing +x)

yunqu: Got it. Will change.

sdbuild/packages/xrt/qemu.sh
```diff
@@ -0,0 +1,36 @@
+# Set up some environment variables as /etc/environment
+# isn't sourced in chroot
+
+set -e
+set -x
+
+# to be added into the main flow
+apt-get update
+apt-get -y install libgtest-dev libboost-all-dev uuid-dev libyaml-dev 
+apt-get -y install opencl-headers ocl-icd-dev  ocl-icd-opencl-dev
```
yunqu: @PeterOgden  For line 8 ~ 10, should we add them into the multistrap? Let me know so I can fix this.

PeterOgden: Seems reasonable to me

PR #898, opened 2019-08-16
Add new mem_dict currently consisting of just PSDDR

PeterOgden:


PeterOgden:
Only the last commit is the PR - the rest is the dependency on #895

schelleg:
added last commit on image_v2.5 branch - now b20b5ca298b2bf8a8789dcddb19be3b04b09d813

PR #906, opened 2019-08-22
build pynq from sdist and add sdist target

yunqu:
This pull request has dependency on pull request #905


sdbuild/Makefile
```diff
@@ -261,7 +266,14 @@ $$(SYSROOT_$1): $$(IMAGE_$1) | $$(SYSROOT_ROOT)
 
 sysroot_$1 : $$(SYSROOT_$1)
 
-.PHONY: sysroot_$1 sdx_sw_$1
+$$(DIST_$1): $$(IMAGE_$1) | $$(DIST_ROOT)
+	@if [[ $1 == "Pynq-Z1" || $1 == "Pynq-Z2" || $1 == "ZCU104" ]]; then \
```
giunatale: Why only for included boards? If I build from scratch for an external board, the tarball is nowhere to be found.

PR #966, opened 2019-10-24
Fix install dependencies

myorbit:
Please consider this fix.
Otherwise the provided Vagrantfile cannot finish successfully as setup_host.sh fails because it is unable to locate docker related packages.

### DIFF
--- a/sdbuild/scripts/setup_host.sh
+++ b/sdbuild/scripts/setup_host.sh
@@ -50,11 +50,25 @@ socat
 zlib1g-dev
 zlib1g:i386
 gcc-multilib
+unzip
 EOT
 set -e
 
 sudo apt-get update
 sudo apt purge -y libgnutls-dev
+
+# Setup docker and containerd using repository before installing them
+sudo apt-get install -y \
+        apt-transport-https \
+        ca-certificates \
+        curl \
+        gnupg-agent \
+        software-properties-common
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+sudo add-apt-repository \
+        "deb [arch=amd64] https://download.docker.com/linux/ubuntu xenial stable"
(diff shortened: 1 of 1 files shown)


sdbuild/scripts/setup_host.sh
```diff
@@ -61,6 +59,13 @@ sudo add-apt-repository -y ppa:ubuntu-toolchain-r/ppa
 sudo apt-get update
 sudo apt-get install -y $PACKAGES
 
+# Install docker and containerd using the repository
+sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
```
yunqu: I think one more package is missing: gnupg-agent

sdbuild/scripts/setup_host.sh
```diff
@@ -61,6 +59,13 @@ sudo add-apt-repository -y ppa:ubuntu-toolchain-r/ppa
...
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu xenial stable"
+sudo apt-get update
+sudo apt install -y docker-ce docker-ce-cli containerd.io
```
yunqu: I think you can remove this line and revert back line 42 ~ 44?

sdbuild/scripts/setup_host.sh
myorbit: I don't think `gnupg-agent` is necessary in this case. It is a daemon 
"that facilitate not having to enter the password for your key every time you want to use it." [1]
[1] https://help.ubuntu.com/community/GnuPrivacyGuardHowto#Tips_and_Tricks

myorbit: Following the official docker install recommendation.

PR #974, opened 2019-12-05
Enable Ubuntu 18.04 host support

powderluv:
Fix some packages that have changed since 16.04.
Remove some supurious checks in checkenv
Fix auth issues in multistrap
Dont rely on fancy resolvconf and just use nameserver 8.8.8.8
(aka google public dns)

Builds end to end on an 18.04 system and should workon on 19.x too


giunatale:
Hi @powderluv, we appreciate your contribution! Thanks for taking the time to put this together.

However, we won't be able to accept it if Google DNS is used in resolv.conf, due to our internal policies.
I understand that from your point of view this makes sense (I have also looked at https://github.com/Avnet/Ultra96-PYNQ/issues/46), but since we use the build flow internally for distributing new images, we need to rely on our corporate network configuration.
We open-sourced the build flow for everyone to use, but since it is used internally, this specific change cannot be merged in.
I would suggest you remove it and instead open a thread on our forum https://discuss.pynq.io documenting your solution.

Also, did you make sure it is still working on 16.04? As of now, this is a strict requirement for us.

powderluv:
No worries. I understand. I have a Work in progress patch that fixes the resolv.conf issue in the chroot without using a hardcoded DNS.  I also have a Dockerfile to make it easy to test on 16.04/18.04. Will cleanup and update this PR.

giunatale:
That's great to hear! please feel free to get in touch in case you need assistance.

powderluv:
I have fixed the resolv.conf issues without the using a hardcoded nameserver. Essentially I copy over the resolv.conf from the host's /run directory. I also fixed [1] and added a way to add different DEPS for different host Ubuntu versions (gcc-multilib doesn't exist in 18.04 and later) 

1) https://discuss.pynq.io/t/pynq-sd-image-for-ultra96-error-in-gpg-key-for-apt/101/3

giunatale:
Thanks @powderluv, I am taking care of testing this and will report out. We appreciate your work!

sdbuild/ubuntu/bionic/aarch64/multistrap.config
```diff
@@ -8,6 +8,7 @@ addimportant=true
 [Bionic]
 source=http://ports.ubuntu.com/ubuntu-ports
 suite=bionic universe
+keyring=ubuntu-keyring
```
ileop: Maybe this could be added to `sdbuild/ubuntu/bionic/arm/multistrap.config`.

I was encountering issues with GPG and not verifiable keys (3B4FE6ACC0B21F32) and your fix solved it.
However, I am working with a customized version of the Pynq-Z2, so I had to apply your changes to arm platforms.

(Working with Virtual Box, Ubuntu 18.04.3 LTS)
--
Btw, thanks for your help.

powderluv: Updated to build for arm boards too. Thx for your feedback.

sdbuild/ubuntu/bionic/aarch64/config
```diff
@@ -1,3 +1,3 @@
-STAGE2_PACKAGES_aarch64 := gcc-mb python_packages_bionic libsds jupyter
+STAGE2_PACKAGES_aarch64 := gcc-mb libsds python_packages_bionic jupyter
```
powderluv: just FYI: libsds provides libxlnk_cma.h which is required by the _xhdmi packages that are built by PYNQ in the python_packages_bionic.

giunatale:
Ok so, first of all, these changes should be integrated in your PR
https://github.com/giunatale/PYNQ/commit/eef551931c95f690970a9a2e0240bd6edb0045fa

As of now, we still are not able to get it to work though. Not on 18.04, nor 16.04
This specific change is not working for us: https://github.com/Xilinx/PYNQ/pull/974/files#diff-93512a360433648fc68944b170b39154R80
`/run/systemd/resolve` does not exists on our Docker images and build systems.

Can you comment on why you did that specific change? We are also wondering why you do that copy in `$target/run/systemd`.

Thanks

giunatale:
I can see why you do that copy in $target that's because if we remove the force-reinstall of resolvconf in `create_rootfs.sh` that's where it is picked from.
Indeed /etc/resolv.conf in chroot is a symlink there
```/etc/resolv.conf -> ../run/systemd/resolve/stub-resolv.conf```

We don't really want to do this copy in chroot, the best approach is to let chroot pick up the right resolv.conf from host, no matter how the sysconfig (again, for instance, in our environments /run/systemd/resolve does not exist). And everything should be done host-side. We'll do some more testing ourselves, but you are welcome to investigate this as well.

powderluv:
Please feel free to break up the changes and merge them in. I dont care about it being my patch as much as getting the build fixed :) 

So  giunatale@eef5519 looks great. Please merge that if that passes all your tests.

powderluv:
About resolv.conf being a dangling reference: I see few solutions

1) use a resolv.conf that hardcoded during chroot setup. This was my initial commit in this PR that uses https://developers.google.com/speed/public-dns  .  Since there is no private elements it could work but I understand the need to use your own DNS.
2) use a resolv.conf that copies over from the host system (when running natively - no vagrant or Docker). This is the latest patch I have 
3) Bind mount /run/systemd/resolv into the chroot in your docker run command. But since the Open Source build doesn't have a Dockerfile I can't submit a patch for it. I do have a docker setup with all the Deps in setup_host.sh. Maybe you can push an opensource Dockerfile and I can add the resolv bind mount into it ?

The other option I could think of is to check my cp of /run/systemd/resolve only if it exists. This will also work for your Dockerfile and for "native" systems. LMK if that is good and I can rework the patch again.

yunqu:
I don't see `setup_host.sh` gets added - maybe it means you use the same `setup_host.sh` for both 16.04 and 18.04?

sdbuild/ubuntu/bionic/aarch64/config
yunqu: `pynq` has its dedicated turn to build; I don't think it is getting built in `python_packages_bionic`?

powderluv:
Good point.  I will update a new setup_host.sh  in a separate patch that essentially installs:

for 18.04 and later
gcc-multilib-arm-linux-gnueabihf
gcc-arm-linux-gnueabihf
g++-arm-linux-gnueabihf
gcc-aarch64-linux-gnu
g++-aarch64-linux-gnu
 
and for 16.04 
gcc-multilib

sdbuild/ubuntu/bionic/aarch64/config
powderluv: is that guaranteed to be before libsds ?  Im running on a 112 core system vs the Vagrant default. Happy to try it out with this change and piecemeal the changes into master.

yunqu: If you look at the same file, you can see `pynq` gets installed later. So what I am saying is that you don't have to change this part.

powderluv: ok. Will remove in next patch. Looking for guidance on resolv.conf

giunatale:
Sorry replying from my phone.
@ynqu setup_host is updated in my branch. I am missing a few packages @powderluv mentioned though, but I recommend we merge the two branches so the changes proposed can be integrated with what I already did.
@powderluv to merge in my changes I either do a new PR or we merge it in your branch. If you want me to do the latter (which I actually recommend we do) you should give me write access on your repo so I can push the changes on that branch. Or you can merge them in yourself. I used your branch as base.

Regarding resolve.conf, unfortunately we cannot publish our Dockerfiles, sorry.
The last option is actually what I would do indeed. /etc/resolve.conf will always exist and is going to be a symlink to the used resolve.conf, so we could follow that symlink and find the used resolve.conf on host that way, and copy that instead of assuming it is in a fixed place, that might change based on the os config

powderluv:
no worries. I gave you push access to my repo. Feel free to update the branch there. 

Once your changes are in I can rework resolv.conf with a check.

giunatale:
That's great. I pushed my changes :)

powderluv:
Can you push it to a new branch? We can pull your change in first and then bring in the resolve.conf change since that needs updating. Thanks

giunatale:
Pushed only my commit (I rebased onto Xilinx:master) in https://github.com/powderluv/PYNQ/tree/18.04-noauth-setup_host

I hope that's what you were asking, if not let me know so I can act on it.

giunatale:
> gcc-multilib-arm-linux-gnueabihf
> gcc-arm-linux-gnueabihf
> g++-arm-linux-gnueabihf
> gcc-aarch64-linux-gnu
> g++-aarch64-linux-gnu

FYI, I force-pushed this change on the `18.04-noauth-setup_host` branch I created on your repo.

giunatale:
Ok so, I've been looking at it a bit more in depth and I think I've solved the issue in a way that should satisfy everyone.
https://github.com/giunatale/PYNQ/commit/be6aa8a7a9425b8ef9aae0eb9124f766214275b0
I've fixed it in setup_host.sh (so host-side) and this is also because in the end, your copy was not really affecting the chroot env. This is because when we install packages, we mount sys so your copy gets blown away. The true fix for you was preventing the force-reinstall of resolveconf which is making chroot point at the right resolve.conf file (the one you have on your host).

There's another change that we require to do for our internal build that you can see in `create_rootfs.sh` along with the removal of the resolv.conf copy in `$target`, but this I think all we need.
Along with my changes on the branch I pushed on your repo, this I think should be it.
If you want, I can merge everything in this PR branch so we should be able to accept the PR. Indeed, my current tests seems to show that everything is working fine. I'll do more testing but for now this is promising!
Of course, you are strongly invited to test this yourself and provide feedback.

Please advise on how you want to proceed. And many thanks for this PR!

Cheers!

powderluv:
This is great. I’ll retest in the next few hours and report back. I can update this PR with all changes so it can be merged

powderluv:
Btw if you have access to the team that ships petalinux can you please get them to update check_env binary that gets called from settings.sh to not check for gcc-multilib on 18.04+ systems since the package name has changed. It is just a harmless warning.

powderluv:
I updated this PR with the three commits from  https://github.com/giunatale/PYNQ/commit/be6aa8a7a9425b8ef9aae0eb9124f766214275b0 so hopefully it is ready to be merged in.

giunatale:
> Btw if you have access to the team that ships petalinux can you please get them to update check_env binary that gets called from settings.sh to not check for gcc-multilib on 18.04+ systems since the package name has changed. It is just a harmless warning.

Yes I noiced the same. I'll see what we can do. Fortunately it doesn't really affect anything, as you said it's just a harmless warning.

> I updated this PR with the three commits from [giunatale@be6aa8a](https://github.com/giunatale/PYNQ/commit/be6aa8a7a9425b8ef9aae0eb9124f766214275b0) so hopefully it is ready to be merged in.

I had to add a small thing on setup_host.sh and check_env.sh (I hope you don't mind) which is install/check for `rsync`. This is actually a package that should always be pre-installed, but on my 18.04 docker it wasn't there and Petalinux failed for that reason.
That said, if you confirm that this is working for you (for me it is on all my tests) I can notify the team so it could be merged in. (I also take care of addressing your comments, thanks for taking the time to review them)

powderluv:
Works for me. Please merge

giunatale:
We are doing some more testing, we'll merge once we are 100% sure it's working in all our test cases, which should happen soon. In the meantime, I've pushed a small change in setup_host and check_env (python3-pip as a dep)

powderluv:
Thanks for the update. Will let you update it as required and merge.

yunqu:
@powderluv  can I put PR onto your gihub repo? Right now I don't have access. I need to fix the vagrant file and a couple of other places.

powderluv:
Done

powderluv:
Any updates on the testing ? Please merge if it looks ok.

yunqu:
For bionic VM, `xterm` and `gcc-multilib` are still required to install petalinux. Please give me a few days to further verify this.

yunqu:
> for 18.04 and later
> gcc-multilib-arm-linux-gnueabihf
> gcc-arm-linux-gnueabihf
> g++-arm-linux-gnueabihf
> gcc-aarch64-linux-gnu
> g++-aarch64-linux-gnu

@powderluv @giunatale I found that this does not work for vagrant. Petalinux still requires `gcc-multilib` to install. Looks like `gcc-multilib` and `gcc-multilib-arm-linux-gnueabihf` cannot coexist at the same time. Workaround would be use `gcc-multilib` first to make Petalinux happy, then install `gcc-multilib-arm-linux-gnueabihf` after Petalinux has been installed. This is ugly, but if you guys have better idea, let me know.

powderluv:
I think that is a reasonable workaround. We have to do that even in native 18.04 and above systems. I would suggest a fix to petalinux instead.

powderluv:
Btw once we merge this If anyone is interested I can have a follow on PR  that will fix the build for 19.10 and upcoming 20.04 but it requires some bb fixes to qemu until they update the yocto SDK

giunatale:
> > for 18.04 and later
> > gcc-multilib-arm-linux-gnueabihf
> > gcc-arm-linux-gnueabihf
> > g++-arm-linux-gnueabihf
> > gcc-aarch64-linux-gnu
> > g++-aarch64-linux-gnu
> 
> @powderluv @giunatale I found that this does not work for vagrant. Petalinux still requires `gcc-multilib` to install. Looks like `gcc-multilib` and `gcc-multilib-arm-linux-gnueabihf` cannot coexist at the same time. Workaround would be use `gcc-multilib` first to make Petalinux happy, then install `gcc-multilib-arm-linux-gnueabihf` after Petalinux has been installed. This is ugly, but if you guys have better idea, let me know.

I agree with @powderluv, it sounds reasonable to me. Although I also agree this should be fixed on the Petalinux side. The problem is @powderluv we have little control over that. So for now, I'd follow @yunqu strategy.

powderluv:
> @giunatale Is there a reason why we want to have PACKAGES_18_04? I found that with gcc-multilib I am able to build the SD card.

This sets us up for the future. Anything after 18_04 will also require ncurses6 instead of ncurses5 (since that has been removed in 19.XX+ )

powderluv:
> I agree with @powderluv, it sounds reasonable to me. Although I also agree this should be fixed on the Petalinux side. The problem is @powderluv we have little control over that. So for now, I'd follow @yunqu strategy.

An external contributor will have even less influence on petalinux I assume :) But hopefully there is a feedback channel for you.

yunqu:
> > @giunatale Is there a reason why we want to have PACKAGES_18_04? I found that with gcc-multilib I am able to build the SD card.
> 
> This sets us up for the future. Anything after 18_04 will also require ncurses6 instead of ncurses5 (since that has been removed in 19.XX+ )

My suggestion on this, is to remove PACKAGES_18_04. The problem is that petalinux only supports 16 and 18, based on user guide currently. The PACKAGES_18_04 is not necessary since this may need to be updated again based on new petalinux requirement.

giunatale:
> > > @giunatale Is there a reason why we want to have PACKAGES_18_04? I found that with gcc-multilib I am able to build the SD card.
> > 
> > 
> > This sets us up for the future. Anything after 18_04 will also require ncurses6 instead of ncurses5 (since that has been removed in 19.XX+ )
> 
> My suggestion on this, is to remove PACKAGES_18_04. The problem is that petalinux only supports 16 and 18, based on user guide currently. The PACKAGES_18_04 is not necessary since this may need to be updated again based on new petalinux requirement.

But what @powderluv noted still stands (for ncurses5 vs ncurses6). I am fine removing it for now in setup_host.sh, but it might be required again (or something similar) in the future.

powderluv:
Yung I'm ok with whatever works to merge thIs now.

PR #985, opened 2020-01-09
change overlay.bin to overlay.bit.bin in firmware

yunqu:


PeterOgden:
This seems more like a 2.6 than a 2.5.1 thing given that we're not planning on putting out or testing new images for any of the boards.

Issue #989, opened 2020-01-22
18.04 host support still broken


powderluv:
This is a follow up to pull request #974 

We removed the check for 18.04 packages at the last minute when we merged #974 so if you build on a native host you will get an error. 

If you install gcc-multilib as petalinux suggests and try to build you will get this error:

foo@bar:~/github/PYNQ/sdbuild$ time make BOARDDIR=/home/foo/github/Ultra96-PYNQ BOARDS=Ultra96
/opt/qemu/bin/qemu-aarch64-static -version | fgrep 4.0.0
qemu-aarch64 version 4.0.0                                     
vivado -version | fgrep 2019.1   
Vivado v2019.1 (64-bit)                                             
xsdk -version | fgrep 2019.1    
****** SDK v2019.1 (64-bit)                                                                                                                
which petalinux-config                                                                                                        
/home/foo/peta/tools/common/petalinux/bin/petalinux-config
which arm-linux-gnueabihf-gcc                                           
Makefile:342: recipe for target 'checkenv' failed
make: *** [checkenv] Error 1                                      
                                                                        
real    0m2.204s                                                                            
user    0m2.036s                                                      
sys     0m0.242s                    

If you install gcc-arm-linux-gnueabihf you will uninstall gcc-multilib above and fail:

You will get this error:

libsdl1.2-dev
rsync
python3-pip
gcc-multilib
Error: Package not found - gcc-multilib
Run setup_host.sh
Makefile:342: recipe for target 'checkenv' failed
make: *** [checkenv] Error 1

It is a trivial patch and I'm happy to resubmit it but need your guidance @yunqu @giunatale since I don't want to submit a PR that has no chance of getting to master.

yunqu:
Hi, I have run the following:

```
which arm-linux-gnueabihf-gcc
```
And it gives me:

`/sdbuild/Xilinx/SDK/2019.1/gnu/aarch32/lin/gcc-arm-linux-gnueabi/bin/arm-linux-gnueabihf-gcc`

So I guess the reason why you think the current branch is broken is that you have not installed Xilinx SDK, or forget to source SDK settings?

powderluv:
I have sourced it and all tools including xsdk are in the path. Please see output below 

/opt/qemu/bin/qemu-aarch64-static -version | fgrep 4.0.0
qemu-aarch64 version 4.0.0
vivado -version | fgrep 2019.1
Vivado v2019.1 (64-bit)
xsdk -version | fgrep 2019.1
[...]

PR #990, opened 2020-01-22
Fix 18.04 native builds

powderluv:
Reintroduce the per ubuntu host package version check.

### DIFF
--- a/sdbuild/scripts/check_env.sh
+++ b/sdbuild/scripts/check_env.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Check for dependencies 
-read -d '' DEPS <<EOT
+read -d '' DEPS_COMMON <<EOT
 bc
 gperf
 bison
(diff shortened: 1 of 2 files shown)


powderluv:
Fixes #989

yunqu:
Hi @powderluv , if you install `gcc-multilib-arm-linux-gnueabihf` instead of `gcc-multilib` in `setup_host.sh`, you will not be able to install petalinux directly. You will still have to install `gcc-multilib` first since `gcc-multilib` is a direct requirement for petalinux. So your flow looks like this:

1. install `gcc-multilib-arm-linux-gnueabihf` first
2. install `gcc-multilib`, which overwrites the 1st step.
3. install petalinux.
4. install back `gcc-multilib-arm-linux-gnueabihf` for make to work.

This looks too much. I did not see problem after installing `gcc-multilib` only, but if you do see a problem, I think we need to adjust the check_env, instead of playing with those packages?

powderluv:
I think step one is not required. You can start at step 2. But yeah it is a pain because of petalinux requiring gcc-multilib.  Is there any forum we external folks can file a report on. I'll post on the forums. 

You can't just adjust just check_env since in the Makefile we are checking for 

checkenv: $(patsubst %, qemu_check_%, $(USED_ARCH))
        vivado -version | fgrep ${KERNEL_VERSION}
        xsdk -version | fgrep ${KERNEL_VERSION}
        which petalinux-config
        which arm-linux-gnueabihf-gcc


So you have to either modify Makefile or setup_host .  Either is fine.

powderluv:
btw xsdk is gone in 2019.2 so we have to update this Makefile anyway soon.

yunqu:
@powderluv I meant to change the checkenv in the makefile. But let me investigate this a little bit. I will try again tonight with fresh VMs. From what I have remembered, I did not see any problem last time.

yunqu:
> Reintroduce the per ubuntu host package version check.

@powderluv  At which step exactly is the build process failing?

powderluv:
Please see https://github.com/Xilinx/PYNQ/issues/989 for exact step. It is the make command

PR #1013, opened 2020-02-19
DMA: Enhance DMA driver functionality:

speachy:
* Correct max transfer length (2^sg_length_width - 1)
 * Require use of HWH for attribute detection
 * Ensure data is properly aligned
 * Support hardware configured in scatter-gather or microDMA modes
 * Record/report actual transferred data length
 * Allow sending subset of DMA buffer
 * 64-bit address support

To-do:
 * Error detection & reporting in async/IRQ mode.
 * Scatter-gather support for multiple buffers (requires change to driver API)
 * SG support with MicroDMA (IMO doesn't make a lot of sense to configure, but it is legal)

### DIFF
--- a/pynq/lib/dma.py
+++ b/pynq/lib/dma.py
@@ -28,8 +28,9 @@
 #   ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from pynq import DefaultIP
+from pynq import Xlnk
+import numpy
 import warnings
-
 
 __author__ = 'Peter Ogden, Anurag Dubey'
 __copyright__ = 'Copyright 2017, Xilinx'
@@ -100,7 +101,7 @@ class _SDMAChannel:
 
         """
         if self._interrupt:
-            self._mmio.write(self._offset, 0x1001)
+            self._mmio.write(self._offset, 0x1001) # XXX what about errors?
         else:
             self._mmio.write(self._offset, 0x0001)
         while not self.running:
@@ -116,7 +117,7 @@ class _SDMAChannel:
             pass
 
     def _clear_interrupt(self):
-        self._mmio.write(self._offset + 4, 0x1000)
+        self._mmio.write(self._offset + 4, 0x1000) # XXX what about errors?
 
     def transfer(self, array, start=0, nbytes=0):
         """Transfer memory with the DMA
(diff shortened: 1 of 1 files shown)


schelleg:
Thanks @speachy for the PR... we're almost done wrapping up a release right now and will look to pull this on next week at the latest if no issues.

speachy:
By next week I'll definitely have more to add to this PR.  And hopefully some additional PRs too.  :)

pynq/lib/dma.py
```diff
@@ -101,7 +110,7 @@ def stop(self):
     def _clear_interrupt(self):
         self._mmio.write(self._offset + 4, 0x1000)
 
-    def transfer(self, array):
+    def transfer(self, array, start=0, nbytes=0):
```
PeterOgden: With PYNQ 2.5 we added support for getting valid sub-buffers with the standard `array[start:start+nbytes]` syntax. I'm wary of having multiple ways to accomplish the same thing but happy if there are advantages to a DMA-specific API.

pynq/lib/dma.py
```diff
@@ -110,20 +119,30 @@ def transfer(self, array):
...
             raise RuntimeError('DMA channel not started')
         if not self.idle and not self._first_transfer:
             raise RuntimeError('DMA channel not idle')
+        if nbytes == 0:
+            nbytes = array.nbytes - start
+        if (nbytes > self._size):
+            raise ValueError('Transfer size is {} bytes, which exceeds '
+                             'the maximum DMA buffer size {}.'.format(
+                              nbytes, self._size))
+        if self._dre and ((array.pyhsical_address + start) % 0x40) != 0:
```
PeterOgden: Checking the data-sheet indicates that the alignment for non-SG ports is dependent on the AXI Master width rather than being fixed to 64 bytes.

sdbuild/scripts/create_bsp.sh
```diff
@@ -6,6 +6,14 @@ set -e
 board=$1
 template=$2
 
+# Allow fetching the BSP from a URL (useful for CI integration)
```
PeterOgden: If possible I'd split this out into a separate PR so that we get a clearer history in the repository

pynq/lib/dma.py
speachy: I don't recall seeing that array sub-buffer syntax documented (or used) anywhere.  Not that I tried too hard, as the target platform I'm using will be stuck on v2018.3 for at least another couple of months.

pynq/lib/dma.py
speachy: Hmm, you're right.. "source address must be [MM2S|S2MM] Memory Map data width aligned".  So I guess that means the "c_m_axi_[s2mm|mm2s]_data_width" property?

(And it looks like I got the polarity of the 'self._dre' test backwards.  oops..)

sdbuild/scripts/create_bsp.sh
speachy: Yeah, I hadn't intended to include this in the PR but github automagically included it as I pushed my working tree.  I'll split it out.

yunqu:
+1 with @PeterOgden that we should enforce HWH file.

speachy:
I'll have an updated PR in the next couple of days; still trying to work out the kinks on the SG side.

Given that a great deal of behaviour and sanity checking I've put into the driver is derived from HWH-queried attributes, unless all of that information can also be extracted from the tcl file then the DMA driver can silently fail, or worse, misbehave depending on the HW DMA channel was instantiated.

Just my $0.02..

schelleg:
We will plan on deprecating Tcl-to-metadata in the next official release.  A nice warning/fail in the DMA driver would be helpful to users.

PeterOgden:
@speachy Are you planning on implementing SG support or just having the driver fail to load if it is present? I was contemplating adding a mechanism where a driver could raise an `UnsupportedConfiguration` exception or similar and have the PYNQ framework fall-back to the `DefaultIP` driver. Given that at least I have a couple of examples that use the SG DMA it might be worth rolling those changes in as part of this PR.

speachy:
@PeterOgden Yesterday I updated the commits to include full SG support.  It hasn't seen a lot of testing yet (especially the corner cases) and the way it handles large buffers probably needs revisiting.

The main PR text is out of date but the actual commits have far more detail in them.

speachy:
I updated the main PR text.  Should be up-to-date now.

It is worth noting that the dmachannel.transfer() API still only allows one buffer to be passed in, so full use of SG isn't possible.  However, even in its current form this allows for transfers far larger than the hardware's configured sg_width, as the driver splits it across multiple SG descriptors.

And I'd appreciate guidance on how to go about handling errors in the async/irq mode, as that would require waiting on more than one IRQ at a time...

speachy:
Fixed two more bugs that came up as some of my colleagues banged on this code.  Any further comments?

PeterOgden:
I've just changed the base to a new branch for our next release. The changes look good but we haven't been able to test against all of our designs due to limited lab access at the moment and this way we'll get a full testing cycle in before it lands on master. If you're happy with the I'll accept the PR.

speachy:
There are still some TODOs that I could use some help/feedback with, but I'm happy with what's already there.

speachy:
Any further progress on this, or feedback on how to approach the TODOs?

schelleg:
hi @speachy - let's pull it on this dev branch now and start the testing.  This forces us to test more configs with the DMA.

For TODOs, I see you are catching micro dma in exceptions, so that is a good starting point.  For IRQs and multiple buffers, let's talk more offline to better understand the use case.  As you say, its tricky to get right (and test!).

PR #1022, opened 2020-04-14
PYNQ:  Allow fetching BSP from a URL instead of only local files.

speachy:
This means one can keep the (usually pretty large) BSP separate from the
repository, and has the added benefit of making some CI flows a lot
simpler

### DIFF
--- a/sdbuild/scripts/create_bsp.sh
+++ b/sdbuild/scripts/create_bsp.sh
@@ -6,7 +6,13 @@ set -e
 board=$1
 template=$2
 
-if [ ! -z "$BSP" ]; then
+if [ -n "$BSP" ]; then
+	# If $BSP is a URL, fetch it!
+	if [[ "$BSP" == *"://"* ]] ; then
+		BSP_ABS="${BSP_BUILD}/../downloaded-${BSP_PROJECT}.bsp"
+		curl -o "${BSP_ABS}" "${BSP}"
+		BSP=$(basename "${BSP_ABS}")
+	fi
 	cp -f $BSP_ABS $BSP_BUILD
 	cd $BSP_BUILD
 	old_project=$(echo $(tar -xvf $BSP) | cut -f1 -d" ")
@@ -36,4 +42,3 @@ else
 	petalinux-package --force --bsp -p $BSP_PROJECT \
 		--output $BSP_PROJECT.bsp
 fi
-
--- a/sdbuild/scripts/setup_host.sh
+++ b/sdbuild/scripts/setup_host.sh
@@ -39,6 +39,7 @@ nfs-common
 zerofree
 u-boot-tools
 rpm2cpio
+curl
 docker-ce
 docker-ce-cli
 containerd.io


sdbuild/scripts/create_bsp.sh
```diff
@@ -6,6 +6,13 @@ set -e
 board=$1
 template=$2
 
+# If $BSP is a URL, fetch it!
+if [[ "$BSP" == *"://"* ]] ; then
+       BSP_NEW=`basename "$BSP"`
+       BSP_ABS=$PWD/build/$BSP_NEW
```
giunatale, line 12: I would refrain from re-defining $BSP_ABS in this way. Better to use known env variables defined in the root Makefile
I would rewrite it as
`BSP_ABS=$BSP_BUILD/../$BSP_NEW`

$BSP_BUILD when launching `create_bsp` will be defined as `$$(BUILD_ROOT_$1)/petalinux_bsp` (see [here](https://github.com/Xilinx/PYNQ/blob/master/sdbuild/Makefile#L98)), so moving a level up will provide the location of the build root for the target board.

sdbuild/scripts/create_bsp.sh
```diff
@@ -6,6 +6,13 @@ set -e
 board=$1
 template=$2
 
+# If $BSP is a URL, fetch it!
+if [[ "$BSP" == *"://"* ]] ; then
```
giunatale, line 10: Can you move this if block inside the `if [ ! -z "$BSP" ]` block? It would not be executed anyway if BSP is empty, so it makes things more readable.

sdbuild/scripts/create_bsp.sh
```diff
@@ -6,6 +6,13 @@ set -e
...
+       BSP_NEW=`basename "$BSP"`
```
giunatale, line 11: This will provide unreliable filename in case a the url doesn't contain the actual filename (e.g. if using a url shortener)
maybe it's worth considering using `$BSP_PROJECT.bsp` (see [here](https://github.com/Xilinx/PYNQ/blob/master/sdbuild/Makefile#L100) and two lines below that) especially if you use a dedicated folder like `BSP_ABS=$BSP_BUILD/../downloaded_bsp` as I am suggesting below.

Won't affect functionality, but definitely readability will be improved: better to have predictable places and names for things, will help looking for stuff in case needed/debugging.

giunatale:
Hi @speachy, I made a few comments on this PR, please review and reply to them. If you disagree with some suggestions, we can totally discuss, I am open to it. 

I didn't test this, but in general it looks a really simple snippet of code, so it shouldn't cause troubles, provided you address my comments. Did you test it yourself?

speachy:
I've been using this little snippet for the better part of a year.  :)

Your comments are generally reasonable, I'll respond to them in turn after I finish cleaning out the inbox..

sdbuild/scripts/create_bsp.sh
speachy: My original thinking was to have the block on the outside as a fallback, but yeah, there's clearly no point.  I'll change this.

sdbuild/scripts/create_bsp.sh
speachy: I've changed BSP_ABS to match what you suggested.

I chose to not create a directory as that opens up many more robustness/reliability issues versus just dropping it in the general BSP directory.

sdbuild/scripts/create_bsp.sh
speachy: I've changed the filename to be set to downloaded-${BSP_PROJECT}.bsp.   I don't see the point in creating a separate directory for it given that the file is download into the throwaway per-board build directory already.

speachy:
New version up.  It makes every requested change except for a separate download directory.

giunatale:
Looks good to me now! thanks for taking the time to address my comments 👍

PR #1024, opened 2020-04-19
update docs to use the modern asyncio.run api

graingert:


PeterOgden:
Thanks for the PR. Our PYNQ images are still based on Ubuntu 18.04 and Python 3.6 so we'd prefer not to accept immediately as that's the most common environment PYNQ is used and doesn't have `asyncio.run`. It's possible that this will change in time for v2.6 so I'll keep the PR open and reassign it to a release branch when we have a clearer roadmap.

schelleg:
no plans right now to upgrade asyncio ... I'll close this for now and can reopen asneeded

PR #1031, opened 2020-05-01
Added volume attribute to ADAU1761

Xerusial:
Added volume attribute to ADAU1761 python interface.
It affects play and bypass and can be set by set_volume.

### DIFF
--- a/pynq/lib/audio.py
+++ b/pynq/lib/audio.py
@@ -360,7 +360,6 @@ class AudioADAU1761(DefaultIP):
         self.iic_index = None
         self.uio_index = None
         self.volume = 57
-        self.configure()
 
     bindto = ['xilinx.com:user:audio_codec_ctrl:1.0']
 
--- a/pynq/lib/audio.py
+++ b/pynq/lib/audio.py
@@ -321,6 +321,8 @@ class AudioADAU1761(DefaultIP):
         The index of the IIC instance in /dev.
     uio_index : int
         The index of the UIO instance in /dev.
+    volume : unsigned
+        The output volume of the ADAU1761 IC.
 
     """
     def __init__(self, description):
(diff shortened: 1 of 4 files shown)


pynq/lib/audio.py
```diff
@@ -321,6 +321,8 @@ class AudioADAU1761(DefaultIP):
         The index of the IIC instance in /dev.
     uio_index : int
         The index of the UIO instance in /dev.
+    volume : unsigned
```
yunqu, line 324: This should be `int` since it should be a python type.

yunqu:
Thanks for the contribution. The change looks good to me. I only have 1 comment as added above. We will test and integrate this into the next release.

pynq/lib/audio.py
```diff
@@ -342,20 +344,22 @@ def __init__(self, description):
...
+                          unsigned int volume,
                           int uio_index, int iic_index);""")
 
         self.buffer = numpy.zeros(0).astype(numpy.int32)
         self.sample_rate = None
         self.sample_len = len(self.buffer)
         self.iic_index = None
         self.uio_index = None
+        self.volume = 57
         self.configure()
```
giunatale, line 363: Please remove this line as it conflicts with changes made in https://github.com/Xilinx/PYNQ/pull/1029. We are getting rid of this call here.

yunqu: I think @giunatale means to remove the line `self.configure()`?

giunatale: Indeed. Sorry I pointed to the wrong line.

Xerusial: Should be resolved now.

giunatale: I am not sure, others might comment and clarify in case I am wrong. Since you merged the commit coming from https://github.com/Xilinx/PYNQ/pull/1029, it will probably give a merge-conflict when trying to merge both PRs. The best option would be that you revert the commit altoghether and just remove the line in question from audio.py in a new commit. Please, also force-push these changes as it will keep history clean.

giunatale:
I see you acted on my comment.
Thanks for working with us!

We'll be working on merging 2.6.0 PRs soon, stay tuned, and thanks again!

pynq/lib/audio.py
Xerusial: As nobody has replied to your question, I did as you requested. But just to clarify. I think I did the right thing. The conflicting changes between PRs should be handled by the maintainers of a repository. If two PRs have conflicts, there should be a merge commit. Otherwise, there will be commits reintroducing the same lines over and over again. Just my opinion.

schelleg:
Merged in with bb0a62ed153ab101ebb1ec735b9fe407557490e5 .  Thanks for the contribution!

PR #1055, opened 2020-06-19
Fix Base Overlay Dict in axigpio.rst

AmgadSalama:
The names of the AXI GPIO controller in the first example @ [https://github.com/Xilinx/PYNQ/blob/master/docs/source/pynq_libraries/axigpio.rst](url)

`led_ip = ol.ip_dict['gpio_leds']`
`switches_ip = ol.ip_dict['gpio_switches']`

are not correct. They should be

`led_ip = ol.ip_dict['leds_gpio']`
`switches_ip = ol.ip_dict['switches_gpio']`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -33,5 +33,7 @@
 *.a binary -crlf
 *.bsp binary
 
-# Populate git commit ID into __init__.py
+# Populate git blob ID into __init__.py
+# This ID will always be tied to a specific release tag
+# since the file will be modified to edit the version
 pynq/__init__.py ident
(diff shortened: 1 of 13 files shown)


giunatale:
Hi @AmgadSalama, can you rebase onto master so that your commit is the only one showing? right now you can see a bunch of other commits, we cannot merge this PR in the current state as it would mess up the git history.
Also, it seems your commit to address the issue presented in the comment is absent.

Thanks in advance

yunqu:
I think we can close this issue, because we have used `gpio_leds` on zcu104, but `leds_gpio` on z1/z2. If really needed, we can open another PR saying the actual name may differ on different boards.

PR #1076, opened 2020-07-27
Parse clock dict for pcie devices.

mariodruiz:


mariodruiz:
Pull request ready to be reviewed. Not sure if we want to show the clock dict nicely.

PR #1084, opened 2020-07-29
add deprecation warning for mmio debug keyword

yunqu:


pynq/mmio.py
```diff
@@ -73,6 +75,11 @@ def __init__(self, base_addr, length=4, device=None):
             The length in bytes; default is 4.
 
         """
+        if debug:
+            message = "Debug mode for MMIO is deprecated. " \
+                      "Set debug=False to avoid this warning."
```
PeterOgden, line 80: Don't we want people to not pass debug at all?

yunqu:
Closing this since we can consolidate this with #1088

PR #1094, opened 2020-08-05
clean up dma code

yunqu:


PeterOgden:
While you're touching the file can you replaces the Xlnk calls with allocate ones?

PR #1113, opened 2020-08-24
Fix Microblaze GCC with updated crosstools

PeterOgden:
* Move to the default of binutils 2.32 to match Vitis 2020.1
 * Remove patches that refer to the GDB part of the binutils source tree
 * Remove dependencies on the sysroot as that is no longer required
 * Build version of arm toolchain compatible with Ubuntu 2018.04 and use in place of Vitis compiler


PeterOgden:
Fixes: Xilinx/PYNQ-Dev#272

PR #1115, opened 2020-08-31
Fix Microblaze addressing with 2020.1

PeterOgden:
Vivado 2020.1 changed the way addressing is handled and broke the way we were
mapping system DRAM into an offset address range in the IOP memory maps. To
make this work there is now a separate IP which isolates the system memory
memory map from the microblaze memory map.


schelleg:
hey @PeterOgden - I tried a merge, but the PYNQ-Z2 Vivado build is failing.  My fault - could you try to redo the commit to work with the RPi addressing fix maintained.

PeterOgden:
@schelleg Done

PR #1124, opened 2020-09-09
boot.py service to run user Python during boot sequence

schelleg:
This will place a boot.py in the boot partition of the SD Card that will always run as part of PYNQ's boot sequence.  

That boot.py file is left empty with two commented out examples that users could run including (1) connecting to WiFi and (2) downloading a custom overlay.

### DIFF
--- a/sdbuild/packages/bootpy/boot.py
+++ b/sdbuild/packages/bootpy/boot.py
@@ -1,0 +1,24 @@
+#! /usr/bin/env python3.6
+
+print("bootpy running...")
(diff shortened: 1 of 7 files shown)


PeterOgden:
Do we want to get rid of boot_leds at the same time? Right now there's a race condition between the two which might not end so well if the user downloads a bitstream in boot.py.

PeterOgden:
Fixes: Xilinx/PYNQ-Dev#283

PR #1126, opened 2020-09-11
Fix minor issues with sd build and replace SDK with Vitis in the docs

mariodruiz:
- Add net-tools package in the dependencies
- Check inotify max_user_watches and increase the value if necessary 
- Replace SDK with Vitis in the documentation
- Include Vitis screenshot in the documentation about building app on MicroBlaze
- Include documentation about HDMI IP license

Closes Xilinx/PYNQ-Dev#279


sdbuild/packages/pybind11/qemu.sh
```diff
@@ -1,7 +1,7 @@
 set -x
 set -e
 
-VERSION=2.5.0
+VERSION=2.6.0
```
yunqu, line 4: Pybind11 2.5 is required by Vitis runtime (VART) and I am not sure if 2.6 is compatible with that. So probably we still need 2.5.

mariodruiz: I thought it was related to PYNQ version.

PR #1134, opened 2020-09-25
Interrupt Improvements

PeterOgden:
* Allow for directly connected interrupts
 * Use a warning rather than an exception when a DefaultIP instance contains an incorrectly connected interrupt.


PeterOgden:
Fixes: Xilinx/PYNQ-Dev#274

PR #1163, opened 2020-10-13
Fix circular buffer to overflow properly

mariodruiz:
## Circular buffer
- Add `size_t channels` as arguments to `cb_init` function
- Include channels in the `circular_buffer` struct
- Move the head pointer based on the number of channels being sampled
- Update c programs that use `cb_init`

## Arduino Analog
- Remove floating point operations from microblaze
- Fix log capacity computation
- Update python class, consolidate raw and floating code into single functions and apply conversion based on an argument
- `get_log` returns a numpy array 
- Update notebook accordingly 

## pmod adc
- Modify MAX_SAMPLES to 1008. That means an integer number of samples for 1, 2 and 3 channels


PeterOgden:
Are we able to keep and deprecate the existing APIs? I'm a bit wary of removing functions without warning in our public APIs.

PeterOgden:
I'm mean implement them in terms of the new improved one not keep the existing code

mariodruiz:
@PeterOgden, do you mean `read_raw` `start_log_raw` `stop_log_raw` in the Arduino_Analog class?

PeterOgden:
Yes

PR #1182, opened 2021-01-29
downgrade pip version to make image building

yunqu:
hotfix to allow image to build from scratch

### DIFF
--- a/sdbuild/packages/python_packages_bionic/qemu.sh
+++ b/sdbuild/packages/python_packages_bionic/qemu.sh
@@ -48,6 +48,5 @@ psutil==5.7.0
 pybind11==2.5.0
 EOT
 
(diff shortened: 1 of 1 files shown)


sdbuild/packages/python_packages_bionic/qemu.sh
```diff
@@ -48,6 +48,5 @@ psutil==5.7.0
 pybind11==2.5.0
 EOT
 
-python3.6 -m pip install -U pip
```
giunatale, line 51: I disagree with removing the line entirely, this would result in a build that is inconsistent with the current 2.6

I'd suggest we do 
```python
python3.6 -m pip install -U pip==20.2.3
```

schelleg:
resolving in later PRs

PR #1188, opened 2021-02-28
Change constructor to use DEFAULT_ value for ext oscillator PS clk rate

FredKellerman:
This makes it easier to modify when porting PYNQ to other boards that don't have the same clock rates.


schelleg:
I took the PS Clock commit in #1220 ...

PR #1195, opened 2021-04-09
Add init value and update_pattern to logictools pattern generator

cathalmccabe:
Add initial value to setup()
Add experimental update_pattern() to allow pattern updates from numerical array

### DIFF
--- a/pynq/lib/logictools/pattern_generator.py
+++ b/pynq/lib/logictools/pattern_generator.py
@@ -218,7 +218,7 @@ class PatternGenerator:
     def setup(self, waveform_dict,
               stimulus_group_name=None, analysis_group_name=None,
               mode='single',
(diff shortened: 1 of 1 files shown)


pynq/lib/logictools/pattern_generator.py
```diff
@@ -294,7 +297,38 @@ def setup(self, waveform_dict,
             self.logictools_controller.pin_map[i] = 'OUTPUT'
         for i in self.analysis_pins:
             self.logictools_controller.pin_map[i] = 'INPUT'
+            
+        # Write the initial value to the output pins
+        if(initial_value is not 0xFFFF):
```
PeterOgden, line 302: This is causing a lint error - would you be able to change the `is not` to `!=`?

cathalmccabe: Updated

PR #1198, opened 2021-04-24
Change ADAU1761 codec chip config to slave mode

trulyspinach:
In the base overlay the audio codec control IP `audio_codec_ctrl_v1.0` is generating I2S clocking signal output to ADAU1761. However the chip was configured into master mode and running at its own clock. This will cause random clicking noise when signals are out of phase.

The config code was updated to set codec chip to slave mode.


trulyspinach:
image

PeterOgden:
Thank you for spotting this. Would you be able to change to target 2.7 and I'll accept?

trulyspinach:
Hi PeterOgden, thanks for your reply!
I've changed the target to `image_v2.7`.

PR #1200, opened 2021-04-29
First steps for 20.04

giunatale:
With the proposed PR stage 1 is able to build succesfully. I couldn't get past the stage2 packages, in particular I am having problems with ``python_packages`` and ``jupyter`` stage 2 packages.


schelleg:
Reworked commits into PR in #1222.   I added giunatale as a co-author there and will close this PR.

PR #1234, opened 2021-08-18
Added RFDC packages

lbrown65:
XRFDC, XRFCLK and XSDFEC no longer in individual board repos. Gen 3 support added.


schelleg:
Thanks Lewis!  Great work

PR #1241, opened 2021-08-28
Fixes in AxiGPIO documentation and code

### LINKED ISSUE #1178: Some errors in the document.
* PYNQ version (e.g. v2.5): 2.5
* Board name (e.g. Pynq-Z1): Pynq-Z2
* Description:  I try to follow the example, however, there may be two mistakes in that. 
* 
When the following code ran, I got a KeyError in Jupyter which is `KeyError: 'gpio_leds'`.

I find that the above document mentions that we should use `[led|switch|button|rgbleds]_gpio`, and I do that. 

Then I find it's ok without errors. So I think these two lines may have some errors. 

I think`gpio_leds` should be modified to `leds_gpio` and `gpio_switches` to `switches_gpio`.

* The code
https://github.com/Xilinx/PYNQ/commit/82da59069aa56f62dbc7dda484b6ff2f805b60cc#diff-9b7aa25cb5b22e220f72782035e943209a204ecac4db6ffab0cc0a588bc920e4R67
https://github.com/Xilinx/PYNQ/commit/82da59069aa56f62dbc7dda484b6ff2f805b60cc#diff-9b7aa25cb5b22e220f72782035e943209a204ecac4db6ffab0cc0a588bc920e4R67

* The document
https://github.com/Xilinx/PYNQ/commit/82da59069aa56f62dbc7dda484b6ff2f805b60cc#diff-9b7aa25cb5b22e220f72782035e943209a204ecac4db6ffab0cc0a588bc920e4R67

mariodruiz:
This fixes #1178

### DIFF
--- a/docs/source/pynq_libraries/axigpio.rst
+++ b/docs/source/pynq_libraries/axigpio.rst
@@ -5,7 +5,8 @@ AxiGPIO
 
 The AxiGPIO class provides methods to read, write, and receive 
 interrupts from external general purpose peripherals such as LEDs, 
-buttons, switches connected to the PL using AXI GPIO controller IP.
+buttons, switches connected to the PL using AXI GPIO controller IP. 
+This class is automatically assigned as a driver to IP of the type AXI GPIO
 
 
 Block Diagram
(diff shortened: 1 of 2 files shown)


mariodruiz:
The documentation states if one trying to read an ‘out’ or write an ‘in’ will cause an error. 

https://pynq.readthedocs.io/en/v2.6.1/pynq_libraries/axigpio.html#block-diagram

This was not implemented commit 34a756038ec18da530888e6028551ca24a04b24d addresses that

PR #1323, opened 2021-11-29
Correct environment variables

dspsandbox:
The Pynq build flow always goes through bitstream generation, microbalze BSPs, binaries... even in case a PYNQ_SDIST is specified. This is because *BOARDDIR* and *DEFAULT_BOARDDIR* variables are not part of the environment when launching /sdbuild/packages/pynq/pre.sh.


schelleg:
Thank you for the contribution - give us some time to verify this as part of next image build release.

schelleg:
we are heavily updating sdbuild, so will close this one out just now w/o accepting ... thank you for contributing.

PR #1331, opened 2022-01-31
Move DFX memory mapped regions to the `mem_dict`

mariodruiz:
Work done in this PR

- Include bram controller in the `ip_dict`
 I think it would be nice if the bram controller is included in the `ip_dict` so the `read `and `write `methods from `DefaultIP` can be used directly. 
- Move DFX memory mapped regions to the mem_dict
 This goes hand in hand with #1352 

This fix Xilinx/PYNQ-DEV#361


mariodruiz:
Move this to draft until we have a better solution for the AXI4-Lite interfaces on partial regions

mariodruiz:
TODO: test with HDL module in IPI

PR #1333, opened 2022-01-31
Use new xbutil switches to get XRT version

mariodruiz:
`xbutil dump` is deprecate since XRT 2.12

Fix Xilinx/PYNQ-Dev#347


mariodruiz:
Moving this to draft as it may be a better way of doing this

mariodruiz:
Closing this issue as it is a cleaner way of doing this #1343

PR #1338, opened 2022-02-10
Release the zocl call by calling `device.close()`

mariodruiz:
- Release the zocl call by calling `device.close()`
- keep `pl_server` alive
- grab XRT handle in `_xrt_download` if the handle is empty

With these changes after a `Overlay.free()` the instance of zocl is released, and by not stopping the pl_server we can download another overlay from the same notebook.


mariodruiz:
Converting to draft to do more checking after this https://github.com/Xilinx/Kria-PYNQ/commit/92da27656d00c2b8843986f30590c4c832b80411

PR #1341, opened 2022-02-21
XSA and Block Design Container support

STFleming:
This PR contains:
* An XsaParser in the utils module can parse and load bitstreams and metadata contained within an XSA file format. 
* A fix so that any XSA/HWH files which contain block design container (BDC) elements will not cause a crash when loading the overlay.
* A build-time script so that XSAs generated from designs that utilise BDCs can have missing register metadata packaged into the XSA.


mariodruiz:
Hi Shane, a few comments 
- can you target branch image_v2.8 instead of master for this PR?
- There are a couple of new files without license header, can you add them?
- Can you keep the character count per line below 79? PEP 8 recommends to limit all lines to a maximum of 79 characters https://www.python.org/dev/peps/pep-0008/#maximum-line-length 
- I've noticed that you use double underscore for internal objects `__`. PEP 8 recommends to use one leading underscore for non-public methods, unless your intention is to mangle the names. https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

pynq/utils.py
```diff
@@ -29,18 +29,24 @@
...
 from distutils.file_util import copy_file
 from distutils.command.build import build as dist_build
 from setuptools.command.build_py import build_py as _build_py
+from xml.dom.minidom import Element
+from xml.etree import ElementTree
+from typing import Union
 
 
-__author__ = "Giuseppe Natale"
+__author__ = "Giuseppe Nataleo, Shane T. Fleming, Geoff Gillett"
```
mariodruiz, line 49: extra `o`

PR #1352, opened 2022-03-21
Assign partial dictionary directly.

mariodruiz:
- Assign partial dictionary directly.
- There is a bug in the way phys_addr was treated. The partial HWH provides the absolute offset address not the relative offset address in respect to the partial region apertures


mariodruiz:
This will be incorporated somewhere else

PR #1354, opened 2022-03-25
Deprecating execute_bo_with_waitlist for XRT >=2.11.

mariodruiz:
Raise a RuntimeError exception when ``execute_bo_with_waitlist`` is called,
this method depends on ``xclExecBufWithWaitList`` which is deprecated
since XRT >= 2.11

Fix Xilinx/PYNQ-DEV#358


schelleg:
after these 3 PR accepts - we'll create a v2.8dev0 source distribution to help datacenter Alveo users

PR #1356, opened 2022-03-30
check to see if y2k22 patch exists

skalade:
Base overlay builds fail for people if you don't have the y2k22 patch applied. We can add a check to see if the patch script exists in their installation, and, if it doesn't, raise a warning. This won't be necessary in future releases with newer Vivado/Vitis versions.


schelleg:
agreed with author to not take this one as later tools will not need this patch and it would be deprecated in next release.

PR #1362, opened 2022-05-12
Fixes to support DFX on BDC

mariodruiz:


mariodruiz:
This will be incorporated somewhere else

PR #1367, opened 2022-06-17
MMIO enhancements to support read and write using different datatype

### LINKED ISSUE #634: mmio.write() does not accept numpy.uint32 as type of data argument
If I call [mmio.write()](https://github.com/Xilinx/PYNQ/blob/325efcede748ee4c4198b9e1fc949fb7dbeb7ccb/pynq/mmio.py#L144) when `data` is an instance of `np.uint32`, the method raises a ValueError. This is funky, because the value of `data` is cast to an `np.uint32` when it is written to the `array` member variable inside the method.

I fixed this in my own code by adding:

```
	if isinstance(data, (np.uint32, np.int32)):
            self._debug('Writing 4 bytes to offset {0:x}: {1:x}',
                        offset, data)
            self.array[idx] = data
```

But you might find a better solution (Hence, a raised issue, and not a PR)

mariodruiz:
Support different datatypes for MMIO read and write.

- Merge write functionality in a single method
- MMIO.read arguments `word_order` and `length` are deprecated
- MMIO.write detects datatype automatically 
- MMIO.read returns the datatype based on the dtype keyword (**kwargs). Eg. `mmio.read(offset,dtype=np.float)`, default is int

Fix #634

### DIFF
--- a/pynq/mmio.py
+++ b/pynq/mmio.py
@@ -95,12 +95,8 @@ class MMIO:
         self.length = length
 
         if self.device.has_capability('MEMORY_MAPPED'):
-            self.read = self.read
-            self.write = self.write_mm
             self.array = self.device.mmap(base_addr, length)
         elif self.device.has_capability('REGISTER_RW'):
-            self.read = self.read
-            self.write = self.write_reg
             self._hook = _AccessHook(self.base_addr, self.device)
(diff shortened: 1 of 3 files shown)


schelleg:
Maintaining MMIO API through next release - closing for now

PR #1372, opened 2022-07-22
Include memory grouping in the key `MBG`. Fix #1370

### LINKED ISSUE #1370: Incorrect ip_dict when using multibank Alveo memories
* PYNQ version (e.g. v2.6): 2.7.1
* Board name (e.g. Pynq-Z2): Alveo U280
* Description:

ip_dict is incorrect when multiple memory banks are attached to one kernel AXIMM interface, e.g. for HBM[0:5] the ip_dict will indicate HBM[5]

mariodruiz:
This PR fixes #1370.

The new key MBG is created for arguments that are connected to multiple banks.

### DIFF
--- a/pynq/pl_server/xclbin_parser.py
+++ b/pynq/pl_server/xclbin_parser.py
@@ -176,10 +176,13 @@ def _xclxml_to_ip_dict(raw_xml, xclbin_uuid):
 
 def _add_argument_memory(ip_dict, ip_data, connections, memories):
     import ctypes
-    connection_dict = {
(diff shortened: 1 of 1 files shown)


mariodruiz:
`xclbinutil` source code for the connectivity section is define here https://github.com/Xilinx/XRT/blob/master/src/runtime_src/tools/xclbinutil/SectionConnectivity.cxx and here https://github.com/Xilinx/XRT/blob/master/src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx

PR #1383, opened 2022-09-14
Include driver for AXI4-Stream Switch with AXI4-Lite interface

mariodruiz:
Fix Xilinx/PYNQ-Dev#67


mariodruiz:
This PR needs to be merged first https://github.com/Xilinx/PYNQ/pull/1368

PR #1414, opened 2023-03-08
AddingLMK04828_500.0.txt and LMX2594_4000.0.txt to support RFSoC MTS …

nathanjachimiec:
Adding in LMK04828_500.0.txt and LMX2594_4000.0.txt to support RFSoC MTS design for the ZCU208.  Note that other boards may require TICs export files such as these to support different frequencies and voltage levels.  For instance, the RFSoC4x2 requires HSDS IO standard for its PL and PL_SYSREF signals whereas the ZCU208 uses LVDS.


nathanjachimiec:
better to place with boards instead due to different frequency and IO standards.  Synthesizer configurations are board unique.

Issue #1433, opened 2023-11-16
Exception when an interrupt controller is in a hierarchy


sterin:
* PYNQ version (e.g. v2.6): 3.0.1
* Board name (e.g. Pynq-Z2): Ultra96V2

## Description

When the interrupt controller is in a hierarchy, any attempt to access a component with an interrupt that goes through the interrupt controller results in a `KeyError` exception.

For example, in the following block design, the interrupt controller is named `Controller`, inside a hierarchy called `InterruptController`. 

![image](https://github.com/Xilinx/PYNQ/assets/341250/b2eb2ed7-bda5-4b0c-b034-4f53fc15824e)

When the following Python code is run:

```python
overlay = pynq.Overlay('regfile.bit')
rf_0 = overlay.RegFile_0
```

A `KeyError` exception is thrown:

```python
Traceback (most recent call last):
  File "/home/xilinx/regfile/demo.py", line 5, in <module>
    rf_0 = overlay.RegFile_0
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/pynq/overlay.py", line 363, in __getattr__
    return getattr(self._ip_map, key)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/pynq/overlay.py", line 904, in __getattr__
    driver = ipdescription["driver"](ipdescription)
  File "/home/xilinx/regfile/demo_config.py", line 79, in __init__
    super().__init__(description)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/pynq/overlay.py", line 707, in __init__
    setattr(self, interrupt, Interrupt(details["fullpath"]))
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/pynq/interrupt.py", line 65, in __init__
    parent, self.number = _InterruptController.get_parent(
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/pynq/interrupt.py", line 152, in get_parent
    return _InterruptController.get_controller(parent), number
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/pynq/interrupt.py", line 122, in get_controller
    ret = _InterruptController(name)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/pynq/interrupt.py", line 177, in __init__
    PL.interrupt_controllers[name])
KeyError: 'InterruptController/Controller'
```

## Root cause:

[...]

PR #1434, opened 2023-11-16
Resolves: Exception when an interrupt controller is in a hierarchy

sterin:
(#1433).

Index 'PL.interrupt_controllers' by the 'hierarchy_name' rather than the 'name' of the controller. 

This can be done in several other ways, such as by changing `interrupt_pins` to use the `name` instead of the `hierarchy_name`. I chose one way pretty much arbitrarily.

It would nice if this issue is resolved, either by this PR or by some other way.


ld-cd:
Our lab has hit this too and this patch resolves things for us

STFleming:
Looks great, thanks for the contribution.

PR #1444, opened 2024-04-09
Devcontainer support for SD build

pbk20191:
It would be really nice to have docker container environment for SD build flow. 
Since it will be more lighter than pure virtual machine and get benefit from prebuilt image.

This configuration expects the Xilinx tools all installed in default location `tools/Xilinx` in host. But users can change path if they want to.


skalade:
Would you be willing to turn this into a regular dockerfile, without .devcontainer? We would be more interested in that use-case. This would be a nice alternative to vagrant.

jogomojo:
Closing this as PYNQ v3.1 now has Docker support for SD build flow.

PR #1454, opened 2024-07-03
Update qemu download source

millaker:
When running `setup_host.sh` I found that I cannot access `wiki.qemu-project.org` when downloading qemu 5.2.0 source. The official qemu download source has been changed to `download.qemu.org`.


YangcyXiang:
I meet the same Problem and hope it can be fixed for later users

schelleg:
thank you for the contribution

PR #1471, opened 2025-04-02
Update github workflow

mariodruiz:
- Update Python version matrix 
- Update action versions


schelleg:
Do we want this PR'd onto main branch?

PR #1494, opened 2025-08-06
Parsing I2C Device ID for MIPI PCAM

andrewgschmidt:
Original pcam5c.py used a hardcoded 6 for the /dev/i2c-6 for the PYNQ-ZU and KV260 device.  With the addition of the AUP-ZU3 we need a more generic mechanism to lookup the device index.  This patch looks for the label "RPICAM" to denote the proper offset.

### DIFF
--- a/pynq/lib/video/pcam5c.py
+++ b/pynq/lib/video/pcam5c.py
@@ -4,6 +4,7 @@
 import contextlib
 import os
 from enum import Enum
+import glob
 
 import cffi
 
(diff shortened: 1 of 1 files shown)


pynq/lib/video/pcam5c.py
```diff
@@ -67,10 +68,22 @@ def __init__(self, description):
         virtaddr_gamma_lut = self.gamma_lut.mmio.array.ctypes.data
         virtaddr_demosaic = self.demosaic.mmio.array.ctypes.data
 
-        # todo read /sys/bus/i2c/devices/i2c-6/of_node/label
-
+        i2c_devices = glob.glob('/dev/i2c-*')
+        i2c_index = 6 ## legacy
+        for dev_path in i2c_devices:
+            adapter_number = os.path.basename(dev_path).split('-')[-1]
+            name_path = f'/sys/bus/i2c/devices/i2c-{adapter_number}/of_node/label'
```
mariodruiz, line 75: you could check if the file exists instead of `try`, but this works as well.

mariodruiz (approved):

PR #1499, opened 2025-08-08
fix: update load_ip_data to pull address from base_address instead of addr_range

williamfligor:
Currently in load_ip_data, if the IP exists in the ip_dict the address is invalid as it's made equal to the "addr_range" instead of "base_address"


jogomojo:
Thank you for the PR @williamfligor.

Can I ask how you found this bug? I'm surprised this hasn't come up before.

williamfligor:
> Thank you for the PR @williamfligor.
> 
> Can I ask how you found this bug? I'm surprised this hasn't come up before.

We encountered this bug when trying to use `load_ip_data` to write to memory on a custom `class Custom(Device):`.

PR #1532, opened 2026-01-21
Allow to read mac address from spi flash

akluczyn:
I noticed that since version 3.1 (when petalinux 2024.1 is used) on the Pynq z2 board the MAC address is not retrieved from the flash memory.

This pull request contains patch which again allows the MAC address to be read from SPI. This patch reverts commit ba74bcf from /Xilinx/u-boot-xlnx/ repo and adds changes which was present in patch until version 3.0.1 (commit d75f95a from PYNQ repo)


jogomojo:
Hi @akluczyn 

Have you confirmed this patch works on a Z1/Z2 image? 

IIRC the reason this was removed between PYNQ v3.0 and 3.1 is that the function we used to read from EEPROM had been deprecated. I did look into it for a bit to find an alternative but ran out of time before release.

akluczyn:
Hi @jogomojo,

I have confirmed that it works on Z2 image - I have built an image via docker and after that MAC was correctly retrieved from SPI.

jogomojo:
Thanks for confirming @akluczyn 

I'll test this myself this week and pull it onto the `image_v3.1.3` next release branch

akluczyn:
Hi @jogomojo,

Are there any updates?

PR #1594, opened 2026-08-25
Pin pynqmetadata to pynq-next and move to pydantic 2

lharnaldi:
`import pynq` fails on a freshly built image because the pynqmetadata dependency is unpinned and pip installs the PyPI release (0.1.x), which does not have what this branch needs.

Since #1591 migrated pynq/metadata to pynq-metadata, the shims here import from `pynqmetadata.views.runtime.*`, and PYNQ's Versal support imports `VersalProcSysCore`. Neither exists on PYNQ-Metadata's main branch or on PyPI -- both are on its pynq-next branch (0.2.0). With 0.1.x installed:

    ImportError: cannot import name 'VersalProcSysCore' from 'pynqmetadata'

pynqmetadata 0.2.0 in turn requires pydantic>=2.0, while python_packages_noble pins 1.9.1:

    ImportError: cannot import name 'ConfigDict' from 'pydantic'

so the two changes go together. PYNQ's own code uses only `Field` and `BaseModel`, which exist in both pydantic majors.

Pinning to a branch rather than a tag matches how PYNQ-Metadata is currently published; a released version would be preferable once one exists.

Found while bringing up a VRK160 (Versal Gen 2) board, but neither symptom is board specific.

### DIFF
--- a/sdbuild/packages/python_packages_noble/requirements.txt
+++ b/sdbuild/packages/python_packages_noble/requirements.txt
@@ -84,7 +84,7 @@ ptyprocess==0.7.0
 pure_eval==0.2.3
 pybind11==2.11.1
 pycparser==2.23
-pydantic==1.9.1
+pydantic==2.9.2
 Pygments==2.20.0
 pyparsing==3.3.2
 python-dateutil==2.9.0.post0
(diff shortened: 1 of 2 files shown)


jogomojo:
Hi @lharnaldi 

Thanks for reporting this. The pynq-next branch is our active development branch for the upcoming PYNQ release, so some functionality may be unstable or broken while work is in progress. We expect issues like this to be resolved as part of the release process.

PR #1603, opened 2026-08-27
Fix Versal overlay loading: four defects that only surface in sequence

lharnaldi:
No Versal board can load an overlay from a PYNQ 4.0 image built from
`pynq-next`. This is four separate defects, and they hide one behind the
other: each one stops execution before the next becomes reachable, so
fixing any subset changes nothing observable. All four are needed.

Found while bringing up a VRK-160 (Versal Gen 2), but none of it is
board-specific -- VCK190 is affected identically, and I have tried to keep
every fix general rather than special-casing my board.

### 1. The image ships PYNQ-Metadata 0.1.9

`sdbuild/packages/pynq/qemu.sh` installs PYNQ-Metadata from the default
branch, which is `main` = 0.1.9. That release has no `VersalProcSysCore`,
which `pynq/metadata/clock_dict_view.py` imports, so `import pynq` fails:

    ImportError: cannot import name 'VersalProcSysCore' from 'pynqmetadata'

0.1.9 also declares `pydantic<2`, so pip silently downgrades the pydantic
2.10.6 pinned in `python_packages_noble/requirements.txt` to 1.9.1, leaving
`pydantic_core` 2.23.4 stranded beside it. Nothing in the build log looks
wrong -- pip reports a successful install.

Pinning `pynqmetadata` in `setup.py` does not help here: the sdist is
installed a few lines below with `--no-deps`, so setup.py's requirements are
never resolved in this flow.

### 2. `KeyError: 'divisor1'`

`Overlay.download()` guards on `"divisor0"` being present in `clock_dict`
and then reads `"divisor1"` unconditionally. Versal has a single divisor
stage where Ultrascale has two, so PYNQ-Metadata's `ClockDictView` emits
enable/frequency/divisor0/src_sel and no divisor1.

This is a regression from b5881e0e: PYNQ's own `clock_dict_view.py` gave
Versal neither divisor, so the guard was False and the block was skipped
entirely. The pynq-metadata view adds divisor0 without divisor1, so the
guard now passes and the next line raises.

### 3. Versal detected from the board, not the SoC

`_is_versal_host()` matches `xlnx,versal` in the root `compatible`, which is
the *board* string. VCK190 spells the family into it
("xlnx,versal-vck190-revA") and is detected. VRK-160 gives plain
"xlnx,vrk160" and is not, so `Clocks` falls through to `_ClocksUltrascale`
and drives CRL_APB/CRF_APB registers Versal does not have:

    ValueError: Frequency divider 1 value out of range.

Now checks SoC-level markers, most reliable first: the PLM firmware node
(compatible `xlnx,versal-firmware`, present on every Versal device tree),
then the old root compatible, then the root `family` property.

### 4. `GPIO.get_gpio_base()` returns None

`get_gpio_base_path()` searched for a gpiochip labelled `zynqmp_gpio` or
`zynq_gpio`. Versal labels its PS GPIO `versal_gpio`. Added it; `pmc_gpio`
is deliberately left out, since it is a different controller and
`target_label` already selects it explicitly.

### Verification

On a VRK-160 booting an image built from this branch, with no manual pip
install: `pynqmetadata` 0.2.0, `pydantic` 2.x, `import pynq` clean, and
`BaseOverlay('base.pdi')` loading with GPIO, AXI BRAM and AXI DMA loopback
all exercised through PYNQ.

Note that on Versal the clock call is a no-op by design:
`_ClocksVersal.set_pl_clk` warns that PL clocks are fixed by the boot PDI.
The point of fixes 2 and 3 is to reach that warning instead of an exception.

### DIFF
--- a/pynq/gpio.py
+++ b/pynq/gpio.py
@@ -244,6 +244,10 @@ class GPIO:
         else:
             valid_labels.append('zynqmp_gpio')
             valid_labels.append('zynq_gpio')
+            # Versal's PS GPIO, the analogue of zynqmp_gpio. The PMC also
+            # exposes a 'pmc_gpio' controller; pass it as target_label if
+            # that is the one you want.
+            valid_labels.append('versal_gpio')
 
         for root, dirs, files in os.walk('/sys/class/gpio'):
             for name in dirs:
--- a/pynq/overlay.py
+++ b/pynq/overlay.py
@@ -424,7 +424,10 @@ class Overlay(Bitstream):
                 enable = self.clock_dict[i]["enable"]
                 if "divisor0" in self.clock_dict[i]:
                     div0 = self.clock_dict[i]["divisor0"]
-                    div1 = self.clock_dict[i]["divisor1"]
+                    # Versal has a single divisor stage, so PYNQ-Metadata's
+                    # clock_dict carries divisor0 but no divisor1 for a
+                    # VersalProcSysCore. set_pl_clk already accepts div1=None.
+                    div1 = self.clock_dict[i].get("divisor1")
                     if enable:
                         Clocks.set_pl_clk(i, div0, div1)
                     else:
--- a/pynq/ps.py
+++ b/pynq/ps.py
@@ -16,18 +16,41 @@ ON_TARGET = os.path.isfile('/proc/device-tree/chosen/pynq_board')
 DEFAULT_PL_CLK_MHZ = 100.0
 
 
+# Ordered most to least reliable. The PLM firmware node is SoC-level and
+# present on every Versal device tree; the root `family` property comes from
+# Xilinx's device-tree generator.
+_VERSAL_DT_MARKERS = (
+    ('/proc/device-tree/firmware/versal-firmware/compatible',
+     b'xlnx,versal-firmware'),
+    ('/proc/device-tree/compatible', b'xlnx,versal'),
+    ('/proc/device-tree/family', b'Versal'),
+)
+
+
 def _is_versal_host():
     """Return True when the device tree identifies the SoC as Versal.
 
     ZynqMP and Versal both report `aarch64`, so `CPU_ARCH` cannot tell
     them apart.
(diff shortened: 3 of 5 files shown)


jogomojo:
Hi @lharnaldi 

Thank you again for the contribution.

As with the #1604, I'd appreciate it if these fixes could be split into separate PRs, with a clear way to validate or verify each change independently. Since we're close to release, that makes review and testing much easier. I'd also ask that comments be kept to a minimum in the code itself where possible, with the reasoning for changes captured in commit messages instead.

That said, this is a great PR overall. The addition to `_is_versal_host()` is a genuinely good catch that I likely would have missed, as our current focus has mainly been on the VCK190 while bringing up Versal support.

Regarding the change to `pynq/sdbuild/packages/qemu/qemu.sh`, the GitHub URL should not point to `pynq-next`. That's only a temporary change used when building development images before a release. Once the release is merged into `master`, users building their own images should automatically pick up the correct versions of `PYNQ-Metadata` and `PYNQ-Utils` from the default branches, so this change should not be committed.

Thanks again for taking the time to submit these fixes and improvements.

lharnaldi:
Thanks for the review — split as requested. Each change is now its own PR against pynq-next, one commit each, with a before/after validation recipe in the description and the reasoning moved out of code comments into the commit messages.

#_ Detect Versal from the SoC, not the board's compatible string
#_ Recognise Versal's PS GPIO controller label
#_ selftest: drop the non-existent dma.reset() call
#_ sdbuild: do not leave the DHCP server enabled on boards that never configure it
#_ x11: only start the X server where a display output exists
Two changes did not make the split. The divisor1 fix in pynq/overlay.py is obsolete: a0c65c03 switched Overlay.download() to the frequency-based path and removed the divisor code entirely. And the qemu.sh change pointing PYNQ-Metadata at pynq-next is dropped, per your note that it is a pre-release-only change.

Closing this in favour of the split.

PR #1604, opened 2026-08-27
Three image-hygiene fixes: selftest DMA, unconfigured DHCP server, headless X11

lharnaldi:
Independent of #1603 , no overlap.
Three independent, small defects found while validating a PYNQ 4.0 image.
None is Versal-specific; all three affect other boards.

### `dma.reset()` does not exist

`sdbuild/packages/selftest/tests/python/dma_loopback.py` calls `dma.reset()`
before transferring, which raises

    AttributeError: 'DMA' object has no attribute 'reset'

`pynq.lib.dma.DMA` has no `reset()` -- it exposes sendchannel/recvchannel
plus `set_up_tx_channel()`/`set_up_rx_channel()`, and there is no `reset()`
elsewhere in pynq that this could have meant. It is not needed either:
`DMA.__init__` runs both `set_up_*_channel()` methods, so the overlay
download the test performs immediately before already leaves the channels
configured. As far as I can tell this test cannot have passed on any board.

### The DHCP server is enabled but never configured

multistrap pulls `isc-dhcp-server` into every image and Ubuntu's postinst
leaves both units enabled, but `/etc/default/isc-dhcp-server` is only
written by the `usbgadget` and `wpa_ap` packages. Any board whose package
list has neither boots with

    isc-dhcp-server.service   failed
    isc-dhcp-server6.service  failed

Disabled both in `create_rootfs.sh`, alongside the `wpa_supplicant`,
`apport` and `sleep.target` masks that are there for the same reason, and
`usbgadget` re-enables the v4 unit after writing `INTERFACES="usb0"`.
9b2b0191 addressed half of this inside `usbgadget`, which only helps boards
that include `usbgadget` in the first place; that line is left as it is.

`wpa_ap` does not enable its own service by default, so it gets a matching
commented-out line rather than an unconditional enable -- uncommenting one
without the other would give an access point that hands out no leases.

### X11 starts where there is no display

`pynq-x11.service` is enabled unconditionally, so on a board with no display
output it fails at every boot.

The cause is not the driver. `zocl`, XRT's accelerator DRM driver, claims
`/dev/dri/card0` on every PYNQ image and has no CRTC, and `xorg.conf` names
no BusID, so armsoc opens card0 blindly:

    (II)    DriverName is [zocl]
    (II)    bus_id is [axi:zyxclmm_drm]
    (II) Screen:0,  CRTC:-1
    (EE) Screen(s) found, but none have a usable configuration.

Switching drivers would not help, and a condition on `/dev/dri/card*` would
pass on exactly the boards where it must not, since zocl always creates one.
Gated on a DRM connector instead: the kernel creates `card<N>-<CONNECTOR>`
directories only for a card with real outputs, so X starts where there is a
screen to draw on and the unit stays inactive rather than failed elsewhere.
Boards with a display are unaffected.

### DIFF
--- a/sdbuild/packages/selftest/tests/python/dma_loopback.py
+++ b/sdbuild/packages/selftest/tests/python/dma_loopback.py
@@ -16,7 +16,9 @@ def run(p=None):
         dma = base.dma
     except AttributeError:
         raise FailError("dma not in overlay")
-    dma.reset()
+    # No dma.reset() here: pynq.lib.dma.DMA has no such method, and needs
+    # none -- DMA.__init__ runs set_up_tx_channel()/set_up_rx_channel(), so
+    # the channels are configured by the overlay download just above.
     n = 1024
     tx = allocate(shape=(n,), dtype=np.uint32)
     rx = allocate(shape=(n,), dtype=np.uint32)
--- a/sdbuild/packages/usbgadget/qemu.sh
+++ b/sdbuild/packages/usbgadget/qemu.sh
@@ -29,5 +29,9 @@ subnet 192.168.3.0 netmask 255.255.255.0 {
 EOT
 
 systemctl disable isc-dhcp-server6
+# create_rootfs.sh disables both units in the base image; this package has
+# just written INTERFACES="usb0" into /etc/default/isc-dhcp-server, so the
+# v4 one is wanted here.
+systemctl enable isc-dhcp-server
 systemctl enable usbgadget
 systemctl enable serial-getty@ttyGS0
--- a/sdbuild/packages/wpa_ap/qemu.sh
+++ b/sdbuild/packages/wpa_ap/qemu.sh
@@ -27,5 +27,8 @@ iface wlan0 inet dhcp
 	wpa-conf /etc/wpa_supplicant.conf
 EOT
 
-# Uncomment the following line to enable wireless access point by default
+# Uncomment the following lines to enable wireless access point by default.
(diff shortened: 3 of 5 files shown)


jogomojo:
Hi @lharnaldi 

Thank you for the contribution, I really appreciate it!

Would you mind splitting these changes into separate PRs? Since we're very close to release, I need to be able to review and verify each change independently before merging into `pynq-next`. It would also help if you could provide a straightforward way to validate each contribution, allowing me to verify previous behaviour vs new behaviour. 

Additionally, please keep code comments to a minimum where possible. I generally prefer the reasoning behind changes to be captured in commit messages, as this makes future auditing easier while keeping the code easier to read.

Thanks again for the contribution and for helping improve PYNQ.

lharnaldi:
Thanks for the review — split as requested. Each change is now its own PR against pynq-next, one commit each, with a before/after validation recipe in the description and the reasoning moved out of code comments into the commit messages.

#_ Detect Versal from the SoC, not the board's compatible string
#_ Recognise Versal's PS GPIO controller label
#_ selftest: drop the non-existent dma.reset() call
#_ sdbuild: do not leave the DHCP server enabled on boards that never configure it
#_ x11: only start the X server where a display output exists
Two changes did not make the split. The divisor1 fix in pynq/overlay.py is obsolete: a0c65c03 switched Overlay.download() to the frequency-based path and removed the divisor code entirely. And the qemu.sh change pointing PYNQ-Metadata at pynq-next is dropped, per your note that it is a pre-release-only change.

Closing this in favour of the split.

PR #1605, opened 2026-08-28
Make EDF repo init and sync non-interactive for detached builds

jogomojo:
Redirect repo init stdin and pass -q to repo sync so tmux/detached sdbuild builds do not stall on the color prompt


jogomojo:
This is in addition to #1600 which didn't quiet all the `repo` calls.

PR #1616, opened 2026-08-31
Recognise Versal's PS GPIO controller label

lharnaldi:
GPIO.get_gpio_base_path() searched /sys/class/gpio for a gpiochip labelled zynqmp_gpio or zynq_gpio. Versal labels its PS GPIO controller versal_gpio (and the PMC one pmc_gpio), so the walk found nothing and get_gpio_base() returned None on every Versal board. Added versal_gpio to the default labels; pmc_gpio is deliberately left out, since it is a different controller and target_label already selects it explicitly.

The shared selftest checked for the zynqmp_gpio label by name and reported "no zynqmp_gpio gpiochip label" on Versal, which reads as a board fault rather than a hardcoded expectation. It now accepts the PS GPIO label of any of the three families and names the one it found.

Observed on VRK160, which exposes versal_gpio and pmc_gpio. VCK190 is affected identically.

On a VCK190:

cat /sys/class/gpio/gpiochip*/label
python3 -c 'from pynq import GPIO; print(GPIO.get_gpio_base())'

Before: the labels are versal_gpio and pmc_gpio, neither matches the zynqmp_gpio/zynq_gpio list, and get_gpio_base() returns None. After: it returns the base index.

pmc_gpio is deliberately not added: it is a different controller and target_label already selects it explicitly.

The selftest change is related: it reported "no zynqmp_gpio gpiochip label" on Versal, which reads as a board fault rather than a hardcoded expectation.

### DIFF
--- a/pynq/gpio.py
+++ b/pynq/gpio.py
@@ -244,14 +244,17 @@ class GPIO:
         else:
             valid_labels.append('zynqmp_gpio')
             valid_labels.append('zynq_gpio')
+            valid_labels.append('versal_gpio')
+            valid_labels.append('pmc_gpio')
 
-        for root, dirs, files in os.walk('/sys/class/gpio'):
-            for name in dirs:
-                if 'gpiochip' in name:
-                    with open(os.path.join(root, name, "label")) as fd:
-                        label = fd.read().rstrip()
-                    if label in valid_labels:
-                        return os.path.join(root, name)
+        for wanted in valid_labels:
+            for root, dirs, files in os.walk('/sys/class/gpio'):
+                for name in dirs:
+                    if 'gpiochip' in name:
+                        with open(os.path.join(root, name, "label")) as fd:
+                            label = fd.read().rstrip()
+                        if label == wanted:
+                            return os.path.join(root, name)
 
     @staticmethod
     def get_gpio_base(target_label=None):


jogomojo:
This PR is out of date now due to #1635. Additionally, through testing we found on the VCK190 the GPIO chip was called pmc_gpio, not versal_gpio. Are you seeing something different on the VRK160?

lharnaldi:
Yes -- VRK160 shows both. From its selftest run:

    [FAIL] no zynqmp_gpio gpiochip label (labels: versal_gpio pmc_gpio)

so `versal_gpio` and `pmc_gpio` are both present there, while on your VCK190
only `pmc_gpio` is. I'd guess a device-tree difference in whether the PS GPIO
node is enabled, rather than anything board-specific in PYNQ.

I've rebased onto current `pynq-next` and dropped the selftest half of this
PR: #1635 covers it properly, and passing the label from the manifest is a
better answer than the family-guessing I had.

What's left is just `pynq/gpio.py`. With no `target_label`,
`get_gpio_base_path()` matches only `zynqmp_gpio`/`zynq_gpio`, so on either
Versal board it returns None and `GPIO.get_gpio_pin()` evaluates
`None + GPIO_OFFSET + index` and raises TypeError. That's the documented way
to use the class without naming a label, so it seemed worth keeping the
default list working on Versal -- but if you'd rather Versal users always
pass an explicit label, in the spirit of #1635, say so and I'll close this.

One thing that came out of your question: since a board can expose both
labels, matching against the whole set inside a single directory walk returns
whichever `os.walk` reaches first. The rebased commit iterates over the
labels outermost instead, so a board with both gets the same answer every
boot, with the PS GPIO ahead of the PMC one. Behaviour with an explicit
`target_label` is unchanged.

Also: thanks for merging #1615.

